PR #50, opened 2023-01-18
Search with geocode

shubham7109:
## Description
PR to add a new Kotlin sample _"Search with geocode"_ in `Search and Query` category. This sample finds the location for an address or places of interest near a location or within a specific area

## Links and Data
Sample Epic: `runtime/kotlin/issues/1340 `

## What To Review


prupani-7 (commented):
Nice sample, works well on my device! Like the menu item we have added for the Search

Added couple comments requiring minor edits. 

I have a different concern though.
In the `displaySearchResultOnMap` function, I see that for each entry received in the `geocodeResultList` we are creating a graphic. 
If I enter `san diego zoo` in my search,  the suggestions received from the `locatorTask` are multiple and if I select one address say `San diego Zoo, 2920 Zoo Dr, San Diego, CA, 92101, USA` we have 5 pins added on the map.

My concern is the address I selected is a specific address and it should only point to one location on the map. If I enter the same address in google maps, only one pin gets added based on the address.

In my opinion based on my understanding, it could be confusing for the user, because he is adding one complete address and we are returning 5 geocoded addresses

shubham7109:
@prupani-7 Thanks! I've added in a `multipleResults` flag to the `geocodeAddress()` function to locate for an address pin if a suggestion is clicked. The sample will return multiple `geocodeResults` if a keyword is used like "Zoo", "Starbucks", "Gym".

prupani-7:
Looks good to me, Thanks!

prupani-7 (approved):

search-with-geocode/src/main/java/com/esri/arcgismaps/sample/searchwithgeocode/MainActivity.kt
```diff
@@ -0,0 +1,375 @@
...
+    // geocode parameters used to perform a search
+    private val addressGeocodeParameters: GeocodeParameters = GeocodeParameters().apply {
+        // get all attributes names for the geocode results
+        resultAttributeNames.addAll(listOf("PlaceName", "Place_addr"))
+    }
+
+    // create a new Graphics Overlay
+    private val graphicsOverlay: GraphicsOverlay = GraphicsOverlay()
+
+    // create a picture marker symbol
```
hud10837, line 91: This comment seems misplaced, since it's initialized to null

search-with-geocode/src/main/java/com/esri/arcgismaps/sample/searchwithgeocode/MainActivity.kt
```diff
@@ -0,0 +1,375 @@
...
+                // geocode the typed address, search within map's viewpoint as keyword was submitted
+                geocodeAddress(address, true)
+                addressSearchView.clearAndHideKeyboard()
+                return true
+            }
+
+            override fun onQueryTextChange(newText: String): Boolean {
+                // if the newText string isn't empty, get suggestions from the locator task
+                if (newText.isNotEmpty()) {
+                    lifecycleScope.launch {
```
hud10837, line 167: In theory, would it be possible to have a race condition here? Suppose I type two letters, after the first letter it goes to the server to get a suggestion, and after the second goes to the server to get a suggestion. If the first suggestion takes a long time and the second suggestion is faster, could the suggestions list update with the second suggestion then, with no further input, suddenly update to the first suggestion?

Even if it's possible, it may be rare enough that it's not worth complicating the sample more.

hud10837 (changes requested):
Looks great and the sample runs well, just one question

search-with-geocode/src/main/java/com/esri/arcgismaps/sample/searchwithgeocode/MainActivity.kt
shubham7109: In theory, it would be a race condition here. But in practice, the result returned is always the latest text change as in my testing. So, I will leave it as it is. For samples, we try to stick to code that showcases the SDK functionality and general error handling and leave testing the robustness of the app up to developers.

PR #52, opened 2023-01-26
Change API Key to match Java

markjdugger:
Change API_KEY in gradle.properties to apiKey to match Java tutorials

## Description
Java samples repo uses an API key in the `gradle.properties` in `<userhome>/.gradle`, just as Kotlin samples repo does. They do not, however, use the same property name or value. 
The Java samples' property looks like this:

```
apiKey = <actual 100-character API Key here>
```

whereas Kotlin looks  ike this:

```
API_KEY = "<actual 100-character API Key here>"
```
Our new hire on the SDK docs team, Steve Rozic, found it quite confusing to end up  with `gradle.properties` file like this:

```
apiKey = <actual 100-character API Key here>
API_KEY = "<actual 100-character API Key here>"
```
He though it was an error and that one should be deleted. Of course, that would break the samples repo (and tutorials verification project) for the API Key that one deleted. Since the doc team covers for each other's vacations and other absences, it's important that any colleague can fairly easily fill in during those times, even in they are not completely familiar with the specific SDK.

To achieve that goal, I created this PR. It contains small changes in the `build.gradle` file of each tutorial app.

## Links and Data
- Java samples repo: https://github.com/Esri/arcgis-maps-sdk-java-samples

## What To Review
This PR changes the following.
- Changes this code:
   ```
   buildConfigField("String", "API_KEY", API_KEY)
   ``` 
  to this:
  ```
  buildConfigField("String", "API_KEY", "\"$apiKey\"")
  ```
- Updates the README to refer to `apiKey` instead of `API_KEY`.

Please evaluate these proposed changes, verify that they do not break the samples, and let me know if this is something you would contemplate doing. The PR works fine for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Once the project is cloned to disk you can import into Android Studio:
 Accessing Esri location services, including basemaps, routing, and geocoding, requires authentication using either an API Key or an ArcGIS identity:
  1. API key: A permanent key that gives your application access to Esri location services. Visit your [ArcGIS Developers Dashboard](https://developers.arcgis.com/dashboard) to create a new API key or access an existing API key.
  
-The Android samples in this repository have been structured to use an API key, set once, which will run in all samples. Set your API key in the `gradle.properties` file located in the `/.gradle` folder within your home directory (`/Users/<user_name>/.gradle/gradle.properties`). The API_KEY property should contain quotes around the key itself: `API_KEY = "YOUR_API_KEY"`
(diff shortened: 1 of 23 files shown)


markjdugger:
@shubham7109 Please consider this PR a proposal for your cosideration and Kaushik's too.
Thanks

shubham7109:
Thanks, Mark for the helpful suggestion here. This would make cross-development with Java and Android less of a hassle. In the rare case, this is a requirement, the "fix" here could be helpful. But that being said I could also see this causing disruptions in automatic workflows and a manual edit of every user's global `gradle.properties` file when migrating from the Runtime SDK. After discussing this issue with Rama, I feel it would be best to look into prioritizing this in the future with all SDK platforms in mind regarding the usage of API keys. 

I'll use this PR as a reference as instructions for any request to combine the API keys to match Java 👍

PR #54, opened 2023-01-27
Project Geometry

kaushikrw:
## Description
PR to add a new Kotlin sample _"Project Geometry"_ in `Geometry` category. This sample projects a point from one spatial reference to another.

## Links and Data
Sample Epic: `runtime/kotlin/issues/1123`

## What To Review

## How to Test
- Run the sample on the sample viewer or the repo. 
- Tap on the map once loaded, bottom Textview should display the coordinates in current spatial reference and in the new transformed spatial reference.


project-geometry/src/main/java/com/esri/arcgismaps/sample/projectgeometry/MainActivity.kt
```diff
@@ -0,0 +1,158 @@
...
+                showError("Error Loading Map", mapView)
+            }
+        }
+    }
+
+    /**
+     * Handles the SingleTapEvent by drawing a marker and performs a Spatial reference transformation
+     * of the tapped Location using GeometryEngine and displays the result
+     */
+    private fun geoViewTapped(point: Point) {
```
fpearson1111, line 123: Usually in Android programming, event-handling methods like this are prefixed with "on". I suppose it's kind of a legacy thing from when most APIs were callback-based but I think it still helps readability by indicating that this is called somewhere whenever some event happens.
```suggestion
    private fun onGeoViewTapped(point: Point) {
```

project-geometry/src/main/java/com/esri/arcgismaps/sample/projectgeometry/MainActivity.kt
```diff
@@ -0,0 +1,158 @@
...
+        // show the marker where the user tapped on the map
+        graphicsOverlay.graphics.apply {
+            // this removes previously added markers
+            clear()
+            // add a new marker graphic
+            add(markerGraphic.apply {
+                // updates the geometry location to the current tap
+                geometry = point
+            })
+        }
```
fpearson1111, line 133: The nesting here makes this a bit harder to read than it needs to be. I think first moving the marker graphic and then updating the graphics overlay would be clearer
```suggestion
        markerGraphic.geometry = point
        graphicsOverlay.graphics.apply {
            // this removes previously added markers
            clear()
            // add a new marker graphic
            add(markerGraphic)
        }
```
Although technically there's no need to clear and re-add the marker graphic if it's already there. Its position should update automatically when its geometry is changed.

project-geometry/src/main/java/com/esri/arcgismaps/sample/projectgeometry/MainActivity.kt
```diff
@@ -0,0 +1,158 @@
...
+        "${String.format("%.5f", x)}, ${String.format("%.5f", y)}"
+
+    /**
+     * Displays an error onscreen
+     */
+    private fun showError(message: String, view: View) {
+        Log.e(TAG, message)
+        Snackbar.make(view, message, Snackbar.LENGTH_SHORT).show()
+    }
+}
```
fpearson1111, line 158: Needs a newline

project-geometry/src/main/java/com/esri/arcgismaps/sample/projectgeometry/MainActivity.kt
```diff
@@ -0,0 +1,158 @@
...
+            projectedPoint.toDisplayFormat()
+        )
+    }
+
+    /**
+     * Extension function for the Point type that returns
+     * a float-precision formatted string suitable for display
+     */
+    private fun Point.toDisplayFormat() =
+        "${String.format("%.5f", x)}, ${String.format("%.5f", y)}"
```
fpearson1111, line 149: Good idea to use an extension function here :+1: Although I would suggest moving it because extension functions should generally go outside of class scope, as top-level functions in the file. This is recommended because otherwise there are two `this` references in the body of the function: in this case, the one for Point and the one for MainActivity. Since the function only uses the one for Point it's clearer to put the function outside MainActivity to show that it doesn't depend on MainActivity's state. Usually they're placed at the bottom of the file and still marked with `private`.

project-geometry/src/main/java/com/esri/arcgismaps/sample/projectgeometry/MainActivity.kt
```diff
@@ -0,0 +1,158 @@
...
+            map = ArcGISMap(BasemapStyle.ArcGISNavigationNight)
+            // add the graphics overlay to display marker graphics
+            graphicsOverlays.add(graphicsOverlay)
+            // set the default viewpoint to Redlands,CA
+            setViewpoint(Viewpoint(34.058, -117.195, 5e4))
+        }
+
+        lifecycleScope.launch {
+            // check if the map has loaded successfully
+            mapView.map?.load()?.onSuccess {
```
fpearson1111, line 107: These nullability `?` operators can be avoided by storing the map as a local variable above
```kotlin
// (above...)
val map = ArcGISMap(BasemapStyle.ArcGISNavigationNight)
mapView.apply {
    map = map
}

// (here...)
map.load().onSuccess { // ...
```

fpearson1111 (commented):
Nice work on your first sample :) I have a few comments, mainly about code readability, but overall it looks great.

kaushikrw:
@fpearson1111 Thanks, Finlay. Updated the PR with changes.

project-geometry/README.metadata.json
```diff
@@ -0,0 +1,34 @@
...
+    ],
+    "language": "kotlin",
+    "redirect_from": "",
+    "relevant_apis": [
+        "GeometryEngine",
+        "Point",
+        "SpatialReference"
+    ],
+    "snippets": [
+        "/src/main/java/com/esri/arcgismaps/sample/projectgeometry/MainActivity.kt"
```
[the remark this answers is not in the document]
kaushikrw: Just re-ran the generator, looks like the it's being created with the slash.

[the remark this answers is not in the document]
fpearson1111: Ok cool, I guess that is correct then. @shubham7109 Please confirm :)

project-geometry/src/main/java/com/esri/arcgismaps/sample/projectgeometry/MainActivity.kt
```diff
@@ -91,11 +91,13 @@ class MainActivity : AppCompatActivity() {
...
-
+        // create and add a map with a navigation night basemap style
+        val map = ArcGISMap(BasemapStyle.ArcGISNavigationNight)
         // configure mapView assignments
         mapView.apply {
-            // create and add a map with a navigation night basemap style
-            map = ArcGISMap(BasemapStyle.ArcGISNavigationNight)
+            this.map = map
+            // add our marker overlay to the graphics overlay
+            graphicsOverlay.graphics.add(markerGraphic)
```
fpearson1111, line 100: This change makes sense to me but just want to double-check what the behaviour is when the marker graphic has no geometry (i.e. before `onGeoViewTapped` first runs). Does it still appear on the screen or is it invisible?

kaushikrw: Yeah it doesn't appear on the MapView, since the geometry property is initially null.

fpearson1111 (approved):
Still good to get confirmation about the slash thing but otherwise I am happy to approve :+1:

project-geometry/README.metadata.json
```diff
@@ -0,0 +1,34 @@
+{
+    "category": "Geometry",
```
shubham7109, line 2:
```suggestion
    "category": "Edit and Manage Data",
```
This should resolve the GitHub checker warning.

project-geometry/src/main/java/com/esri/arcgismaps/sample/projectgeometry/MainActivity.kt
```diff
@@ -0,0 +1,152 @@
...
+                showError("Error Loading Map", mapView)
+            }
+        }
+    }
+
+    /**
+     * Handles the SingleTapEvent by drawing a marker and performs a Spatial reference transformation
+     * of the tapped Location using GeometryEngine and displays the result
+     */
+    private fun onGeoViewTapped(point: Point) {
```
shubham7109, line 125:
```suggestion
    private suspend fun onGeoViewTapped(point: Point) {
        // set viewpoint to center to the tapped location
        mapView.setViewpointGeometry(point.extent)
```
Add a line to center the viewpoint to the tapped location

shubham7109 (approved):
Great sample @kaushikrw 🥳! Works well and the SDK implementation looks good 👍 
Just a minor comment to get the GitHub checker to work.

kaushikrw:
@shubham7109 Thanks, Shubham! Updated the PR with changes.

PR #55, opened 2023-01-31
Play KML tour

shubham7109:
## Description
PR to add a new Kotlin sample _"Play KML tour"_ in `Visualization` category.

## Links and Data
Sample Epic: `runtime/kotlin/issues/#1762`


prupani-7 (commented):
Great sample! looks good to me.
I had general comments:
1. When `Play` is clicked for first time, there is some delay when the tour began. Is it due to the time it takes to load the kmlDataSet? 
2. The first location's `Esri HQ` pin gets partially hidden by the layout at the bottom. I am using samsung s9 device. 
image

shubham7109:
@prupani-7 Thanks for the review, I'll address the comments here:
1. When `Play` button is clicked the tour animation starts instantly as the KML file has been loaded with the tour status as `TourStatus.Initliazed`. There are a couple of frames when the tour does not move in the beginning, making it seem like it may be loading. You can verify this by viewing the progress of the tour during the beginning. 
2. The first location's `Esri HQ` pin gets partially hidden due to the dataset used. The same behavior is in my device and Runtime as well.

prupani-7 (approved):

play-kml-tour/src/main/java/com/esri/arcgismaps/sample/playkmltour/MainActivity.kt
```diff
@@ -0,0 +1,252 @@
...
+
+                // collect changes in KML tour status
+                collectKmlTourStatus(kmlTour)
+
+                // set the KML tour to the controller
+                kmlTourController.tour = kmlTour
+            }
+        }
+
+        resetTourButton.setOnClickListener { kmlTourController.reset() }
```
changanxian, line 127: Could you also reset the KmlLayer be the initial viewpoint?

changanxian (approved):
LGTM, one comment: When I paused, then reset, I expect that the current view is reset to the initial view also.

play-kml-tour/src/main/java/com/esri/arcgismaps/sample/playkmltour/MainActivity.kt
shubham7109: Is there a way to get the initial viewpoint of a KML tour? I tried using `kmlTourController.tour?.viewpoint` but it's always `null`.

changanxian: one way you could do it is when the `play` is clicked, you save the current viewpoint (if it is null), then `reset` is clicked here, then applying the viewpoint here.

shubham7109: Added a `initialViewpoint` instance to set `SceneView.setViewpoint` when the reset button is clicked

PR #58, opened 2023-02-04
Navigate route

shubham7109:
## Description
PR to add a new Kotlin sample _"Navigate route"_ in `Routing and Logistics` category.

## Links and Data
Sample Epic: `runtime/kotlin/issues/1393`

## What To Review
-  Review the sample to make sure it is easy to follow like other samples on Android
- Review that Kotlin code for suitable sample implementation of the Maps SDK


prupani-7 (approved):
Sample looks good to me! I have a comment about the `sampleserver6 NAservice` we are using, and that we should avoid using that server. Devops is working on providing us same service hosted on `sampleserver7`, I think we should use that here and test this sample once. I am happy to approve this so the review process continues.

navigate-route/src/main/java/com/esri/arcgismaps/sample/navigateroute/MainActivity.kt
```diff
@@ -0,0 +1,405 @@
...
+        lifecycleScope.launch {
+            routeTracker.newVoiceGuidance.collect { voiceGuidance ->
+                // use Android's text to speech to speak the voice guidance
+                speakVoiceGuidance(voiceGuidance.text)
+                nextDirectionTextView.text = getString(
+                    R.string.next_direction,
+                    voiceGuidance.text
+                )
+            }
+        }
```
fpearson1111, line 320: I'm not sure about this because every time a new location is received this method is called which launches a new coroutine to collect voice guidance. But really there only needs to be one collection voice guidance in the whole application. My first thought would be to move this code into `startNavigation` although that also isn't ideal because every time the route is restarted there will be a new coroutine for collecting voice guidance. It's probably ok though.

navigate-route/src/main/java/com/esri/arcgismaps/sample/navigateroute/MainActivity.kt
```diff
@@ -0,0 +1,405 @@
...
+            textToSpeech?.speak(voiceGuidanceText, TextToSpeech.QUEUE_FLUSH, null, null)
+        }
+    }
+
+    private fun showError(message: String) {
+        Log.e(TAG, message)
+        Snackbar.make(mapView, message, Snackbar.LENGTH_SHORT).show()
+    }
+}
+
```
fpearson1111, line 405: Should this sample also take advantage of the `RouteTracker.isSpeechEngineReady` callback somewhere? Maybe it's not strictly needed but I think it makes sense to show off that functionality.

fpearson1111 (commented):
I wasn't able to try out this sample because I got an error about `sampleserver6` not being trusted. I guess this link needs to be updated to `sampleserver7`?

navigate-route/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,14 @@
+<resources>
+    <string name="app_name">Navigate route</string>
+    <string name="routing_service_url">https://sampleserver6.arcgisonline.com/arcgis/rest/services/NetworkAnalysis/SanDiego/NAServer/Route</string>
```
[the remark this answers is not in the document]
shubham7109: Updated to use SS7 👍

navigate-route/src/main/java/com/esri/arcgismaps/sample/navigateroute/MainActivity.kt
shubham7109: Would this implementation be correct? Check if the speech is ready before collecting new `newVoiceGuidance`
```kotlin
// if route tracker speech engine is ready
if (routeTracker.isSpeechEngineReady.invoke()) {
    // listen for new voice guidance events
    lifecycleScope.launch {
        routeTracker.newVoiceGuidance.collect { voiceGuidance ->
            // use Android's text to speech to speak the voice guidance
            speakVoiceGuidance(voiceGuidance.text)
            nextDirectionTextView.text = getString(
                R.string.next_direction,
                voiceGuidance.text
            )
        }
    }
}
```

navigate-route/src/main/java/com/esri/arcgismaps/sample/navigateroute/MainActivity.kt
shubham7109: That's a fair point. I've updated the code so the voice guidance is called from `startNavigation` instead of every location update and moved all the voice guidance code to a fun `updateVoiceGuidance()`

shubham7109:
@fpearson1111 Thanks for the feedback, I've updated the service to use SS7. Ready for review 👍

navigate-route/src/main/java/com/esri/arcgismaps/sample/navigateroute/MainActivity.kt
```diff
@@ -0,0 +1,405 @@
...
+                // switch to the next destination
+                routeTracker.switchToNextDestination().getOrElse {
+                    return@launch showError("Error retrieving next destination: ${it.message}")
+                }
+                // set second stop message
+                nextStopTextView.text = resources.getStringArray(R.array.stop_message)[1]
+            } else {
+                // the final destination has been reached,
+                // stop the simulated location data source
+                simulatedLocationDataSource.stop()
```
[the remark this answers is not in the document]
shubham7109: Not sure if it is a bug but when I was to remove `simulatedLocationDataSource` from here, the sample crashes with the error:
```
Caused by: com.arcgismaps.ArcGISException: Invalid argument.: Location data source stop request callback required.
```

navigate-route/src/main/java/com/esri/arcgismaps/sample/navigateroute/MainActivity.kt
fpearson1111: Not quite, that property is supposed to be set with a lambda that returns true if the speech engine is ready and false if it isn't. So my first instinct would be something like
```kotlin
val routeTracker = // .. create route tracker as usual
// set a lambda which returns true if the speech engine is ready for new guidance
routeTracker.isSpeechEngineReady = { isTextToSpeechInitialized && textToSpeech?.isSpeaking == false }
```

But I see how you are interpreting it and that makes sense too (I was involved in the design PR for it so I'm already familiar with how it's supposed to be used) so I wonder if we should change the design of the callback to make it more intuitive.

navigate-route/src/main/java/com/esri/arcgismaps/sample/navigateroute/MainActivity.kt
```diff
@@ -211,128 +208,123 @@ class MainActivity : AppCompatActivity() {
...
+                    routeTrackerLocationDataSource)
+            }
+        }
+
+        lifecycleScope.launch {
+            // if the user navigates the map view away from the
+            // location display, activate the recenter button
+            locationDisplay.autoPanMode.filter { it == LocationDisplayAutoPanMode.Off }
+                .collect { recenterButton.isEnabled = true }
+        }
```
fpearson1111, line 273: Just tried out the sample, this re-centering feature is pretty neat :)

navigate-route/src/main/java/com/esri/arcgismaps/sample/navigateroute/MainActivity.kt
```diff
@@ -211,128 +208,123 @@ class MainActivity : AppCompatActivity() {
...
+            routeGeometry, simulationParameters
+        )
 
         // set up a RouteTracker for navigation along the calculated route
         val routeTracker = RouteTracker(routeResult, 0, true)
 
+        // plays the direction voice guidance
+        lifecycleScope.launch {
+            updateVoiceGuidance(routeTracker)
+        }
```
fpearson1111, line 228: I am still seeing some problems with this if the user resets the route. I found that when starting the route, waiting a few seconds, and resetting it, I started receiving each piece of guidance twice. I think the job from this coroutine would need to be cancelled every time `startNavigation` is called, and same for the coroutine that updates the location text. Or alternatively is it possible to reset the route without recalculating it and creating a new route tracker?

navigate-route/src/main/java/com/esri/arcgismaps/sample/navigateroute/MainActivity.kt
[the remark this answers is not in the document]
fpearson1111: I see the same thing. It seems like a bug, I think we should discuss it.

navigate-route/src/main/java/com/esri/arcgismaps/sample/navigateroute/MainActivity.kt
shubham7109: Noticed, I think canceling the coroutine would make the best sense here. Added the following code to the click listener: 
```kotlin
resetNavigationButton.setOnClickListener {
    lifecycleScope.launch {
        if (locationDisplayJob.isActive) {
            // stop location data sources
            simulatedLocationDataSource.stop()
            routeTrackerLocationDataSource.stop()
            // cancel the coroutine jobs
            locationDisplayJob.cancelAndJoin()
            autoPanModeJob.cancelAndJoin()
            // start navigation again
            startNavigation(routeResult)
        }
    }
}
```

fpearson1111 (approved):
One small comment but otherwise LGTM

navigate-route/src/main/java/com/esri/arcgismaps/sample/navigateroute/MainActivity.kt
```diff
@@ -261,32 +265,46 @@ class MainActivity : AppCompatActivity() {
...
                 .collect { recenterButton.isEnabled = true }
         }
+
+        // reset the navigation if button is clicked
+        resetNavigationButton.setOnClickListener {
+            lifecycleScope.launch {
+                if (locationDisplayJob.isActive) {
+                    // stop location data sources
+                    simulatedLocationDataSource.stop()
+                    routeTrackerLocationDataSource.stop()
```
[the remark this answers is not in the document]
shubham7109: Tested the scenario, reset button resets the route as expected. 👍

PR #59, opened 2023-02-06
Geodatabase replica from feature service

kaushikrw:
## Description
PR to add a new Kotlin sample _"Generate Geodatabase Replica From Feature Service"_ in `Edit and Manage Data` category.

## Links and Data
Sample Epic: `runtime/kotlin/issues/1802`

## What To Review

## How to Test
- Run the sample on the sample viewer or the repo.
- Zoom to any extent and click the generate button. A red outline will show the extent used. 
- The job's progress is shown while the geodatabase is generated. When complete, the map will reload with only the layers in the geodatabase, clipped to the extent.


kaushikrw:
Sorry about the excessive commits to fix the metadata 😅

generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
```diff
@@ -0,0 +1,256 @@
...
+                        progressDialog.progressBar.progress = value
+                        progressDialog.progressTextView.text = "$value%"
+                    }
+                }
+                // start the generateGeodatabase job
+                start()
+                // if the job completed successfully, get the geodatabase from the result
+                val geodatabase = result().getOrElse {
+                    // show an error and return if job failed
+                    showError("Error fetching geodatabase")
```
shubham7109, line 186:
```suggestion
                    showError("Error fetching geodatabase: ${it.message}")
```
Could add the throwable message to the error

generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
```diff
@@ -0,0 +1,256 @@
...
+                geodatabaseSyncTask.unregisterGeodatabase(geodatabase)
+            }
+        }
+    }
+
+    /**
+     * Loads the geodatabase and renders the feature layers on to the map
+     *
+     * @param geodatabase geodatabase to load
+     * @param map map to display it on
```
shubham7109, line 206: Use of `@param` here

generate-geodatabase-replica-from-feature-service/src/main/res/layout/activity_main.xml
```diff
@@ -0,0 +1,31 @@
...
+
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        tools:context=".MainActivity">
+
+        <com.arcgismaps.mapping.view.MapView
+            android:id="@+id/mapView"
+            android:layout_width="match_parent"
+            android:layout_height="match_parent">
```
shubham7109, line 14:
```suggestion
            android:id="@+id/mapView"
            android:layout_width="match_parent"
            android:layout_height="0dp"
            app:layout_constraintBottom_toTopOf="@+id/generateButton"
            app:layout_constraintEnd_toEndOf="parent"
            app:layout_constraintStart_toStartOf="parent"
            app:layout_constraintTop_toTopOf="parent">
```

MapView is constraint to the parent layout, which is causing it to overlap with the button. Set the MapView with the bottom constraint.

generate-geodatabase-replica-from-feature-service/src/main/res/layout/activity_main.xml
```diff
@@ -0,0 +1,31 @@
...
+
+        </com.arcgismaps.mapping.view.MapView>
+
+        <Button
```
shubham7109, line 18:
```suggestion
        <com.google.android.material.button.MaterialButton
```
Just a suggestion, but would be nice if the sample could use Google's `MaterialButton` and use the style `?attr/materialButtonOutlinedStyle`. You could refer to other samples that use the button.

shubham7109 (changes requested):
Great sample Kaushik! I've added in a couple of feedback comments.

generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    /**
+     * Starts a GeodatabaseSyncTask, runs a GenerateGeodatabaseJob and saves
+     * the geodatabase file into local storage
+     *
+     * @param geodatabaseSyncTask the GeodatabaseSyncTask object
+     * @param map mapView to render the feature layers on
+     * @param extents the selected extents of the mapView
+     */
```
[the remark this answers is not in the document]
kaushikrw: Thanks!

generate-geodatabase-replica-from-feature-service/src/main/res/layout/activity_main.xml
kaushikrw: Thanks, that does look better.

generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
```diff
@@ -0,0 +1,256 @@
...
+            // create a generate geodatabase job
+            geodatabaseSyncTask.createGenerateGeodatabaseJob(defaultParameters, geodatabaseFilePath).run {
+                // create a dialog to show the jobs progress
+                val dialog = createProgressDialog(this).apply {
+                    // show the dialog
+                    show()
+                    // allow it to be cancellable
+                    setCancelable(true)
+                    // disable dismissal when tapped outside the dialog
+                    setCanceledOnTouchOutside(false)
```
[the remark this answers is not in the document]
kaushikrw: `setCanceledOnTouchOutside()` is a member of the `AlerDialog` class. I've also added a `setOnCancelListener` to the builder, since upon testing I noticed that cancelling the dialog by the back button doesn't trigger the `setNegativeButton` listener.

generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
```diff
@@ -0,0 +1,256 @@
...
+            }
+            // enable the generate button and add an onClickListener
+            generateButton.isEnabled = true
+            generateButton.setOnClickListener {
+                mapView.visibleArea?.let { polygon ->
+                    // start the geodatabase generation process
+                    generateGeodatabase(geodatabaseSyncTask, map, polygon.extent)
+                }
+            }
+        }
```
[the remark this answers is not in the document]
kaushikrw: Not sure I quite understand the first part. From my understanding, since `load()` is a suspending call, the scope will pause and wait for it to return. Using `getOrElse`, `onFailure` are just ways to access the value of `Result<T>` that `load()` has returned. [kotlin.Result](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/).

I've been using `getOrElse` to get the value of `Result<T>` in cases where it is something other than `Unit/Any`, otherwise using `onFailure`. Especially to avoid nested `onSuccess()/onFailure()`statements.

You're right about the second part. In fact, that's how I had written it initially. But I coupled the creation of `geodatabaseSyncTask` and `generateButton.setOnClickListener{}` with the successful loads as their scope exits in case of failures. Since there is no _refresh_, the only way to retry is to re-launch the app. I'm happy to refactor this though, if the former is the accepted pattern.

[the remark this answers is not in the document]
shubham7109: Seems like your understanding is correct here. We can use your approach here as the standard approach. 👍 

With the second part, the `generateButton` is disabled by default and is enabled when `geodatabaseSyncTask` is loaded. This way the button cannot be clicked until the task successfully loads. So I feel like it could be coupled outside of the coroutine scope:
```kotlin
// create a geodatabase sync task with the feature service url
// This feature service shows a web map of portland street trees,
// their attributes, as well as related inspection information
val geodatabaseSyncTask = GeodatabaseSyncTask(getString(R.string.feature_server_url))

// set the button click listener
generateButton.setOnClickListener {
    mapView.visibleArea?.let { polygon ->
        // start the geodatabase generation process
        generateGeodatabase(geodatabaseSyncTask, map, polygon.extent)
    }
}

lifecycleScope.launch {
    // show the error and return if map load failed
    map.load().onFailure {
        showError("Unable to load map")
        return@launch
    }
    geodatabaseSyncTask.load().onFailure {
        // if the metadata load fails, show the error and return
        showError("Failed to fetch geodatabase metadata")
        return@launch
    }
    // task is loaded, enable the generate button
    generateButton.isEnabled = true
}
```

generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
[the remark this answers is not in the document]
shubham7109: When removing `setCanceledOnTouchOutside`, the cancel listener does get called when pressing the back button for me and I get the expected message: 
> Error fetching geodatabase: User canceled.

Another option would be to `setCancelable(true)`, so the job can only be canceled on cancel button click and not on tapping outside the dialog or a back button press.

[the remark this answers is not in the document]
kaushikrw: The `onCancelListener` does get called as I added that in addition to a `negativeButtonListener`. I think instead setting the dialog to `setCancelable(false)` works better.

shubham7109 (approved):
Looks good! 👍

generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    // shows the geodatabase loading progress
+    private val progressDialog by lazy {
+        GenerateGeodatabaseDialogLayoutBinding.inflate(layoutInflater)
+    }
+
+    // local file path to the geodatabase
+    private val geodatabaseFilePath by lazy {
+        getExternalFilesDir(null)?.path + getString(R.string.portland_trees_geodatabase_file)
```
hud10837, line 74: This is something I saw recently across samples, it's a nitpick though. I think generally it's probably better to use `File(getExternalFilesDir(null)?, getString(R.string.portland_trees_geodatabase_file)` because it would automatically handle the file separators and you can verify that the file exists etc. However in this case it's probably better to match the general pattern of the samples so there's not one sample using `File` and the rest aren't.

That said, does this line need a separator?
```suggestion
        getExternalFilesDir(null)?.path + File.separator() +  getString(R.string.portland_trees_geodatabase_file)
```

generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
```diff
@@ -0,0 +1,259 @@
...
+            }
+
+            geodatabaseSyncTask.load().onFailure {
+                // if the metadata load fails, show the error and return
+                showError("Failed to fetch geodatabase metadata")
+                return@launch
+            }
+
+            // enable the generate button since the task is now loaded
+            generateButton.isEnabled = true
```
hud10837, line 136: Nice 👍

generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
```diff
@@ -0,0 +1,259 @@
...
+        lifecycleScope.launch {
+            // create generate geodatabase parameters for the selected extents
+            val defaultParameters =
+                geodatabaseSyncTask.createDefaultGenerateGeodatabaseParameters(extents).getOrElse {
+                    // show the error and return if the task fails
+                    showError("Error creating geodatabase parameters")
+                    return@launch
+                }
+
+            // set return attachments option to false
```
hud10837, line 168: It might be worth adding what this means, for users who are not familiar why this should be false.

generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
```diff
@@ -0,0 +1,259 @@
...
+            defaultParameters.returnAttachments = false
+            // create a generate geodatabase job
+            geodatabaseSyncTask.createGenerateGeodatabaseJob(defaultParameters, geodatabaseFilePath)
+                .run {
```
hud10837, line 172: Good reminder to use `run`

generate-geodatabase-replica-from-feature-service/README.md
```diff
@@ -0,0 +1,32 @@
...
+## How to use the sample
+
+Zoom to any extent. Then click the generate button to generate a geodatabase of features from a feature service filtered to the current extent. A red outline will show the extent used. The job's progress is shown while the geodatabase is generated. When complete, the map will reload with only the layers in the geodatabase, clipped to the extent.
+
+## How it works
+
+1. Create a `GeodatabaseSyncTask` with the URL of the feature service and load it.
+2. Create `GenerateGeodatabaseParameters` specifying the extent and whether to include attachments.
+3. Create a `GenerateGeodatabaseJob` with `geodatabaseSyncTask.generateGeodatabaseAsync(parameters, downloadPath)`. Start the job with `job.start()`.
+4. When the job is done, `job.getResult()` will return the geodatabase. Inside the geodatabase are feature tables which can be used to add feature layers to the map.
```
hud10837, line 20: I think this should just be `result()`
```suggestion
4. When the job is done, `job.result()` will return the geodatabase. Inside the geodatabase are feature tables which can be used to add feature layers to the map.
```

hud10837 (approved):
Very kotliny. LGTM, a couple potential changes but nothing that requires a re-review.

generate-geodatabase-replica-from-feature-service/README.md
kaushikrw: Good catch!

generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
kaushikrw: Thanks!

generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
shubham7109: I think @hud10837 is reffering to the call to set the `provisionPath` once the library downloads the portal item where the use of `File.separator` is required:
```kotlin
private val provisionPath: String by lazy {
    getExternalFilesDir(null)?.path.toString() + File.separator + getString(R.string.app_name)
}
```
Since this is not related to the library, in general, it's okay to add the line separator to the string filename instead 👍

PR #60, opened 2023-02-06
Find route around barriers

prupani-7:
## Description
PR to add a new Kotlin sample "Find route around barriers" in Routes and Logistics category. This sample finds a route that reaches all stops without crossing any barriers.

## Links and Data
Sample Epic: runtime/kotlin/issues/1695 

## What To Review


shubham7109:
Hopefully, you don't mind but I've updated the dialog to use view binding instead of the `findViewById<>`. And updated the switches to persist the state of the toggle to match the iOS sample.

shubham7109 (changes requested):
Really impressive work with this sample @prupani-7. This sample is a great app to showcase the capabilities of the Maps SDK when many aspects of the API are interacting together in relation to the Android views and the MapView. I just have a couple of comments, but really well-done sample here.

prupani-7:
> Hopefully, you don't mind but I've updated the dialog to use view binding instead of the `findViewById<>`. And updated the switches to persist the state of the toggle to match the iOS sample.

Thanks, yes it is better to use view binding and also have all switches initialized at the start. 
I refactored the switches code, to remove redundancy while checking `findBestSequenceSwitch.isChecked`. It was done twice. Also set the `isChecked = false` on `firstStopSwitch` and `lastStopSwitch` when `findBestSequenceSwitch` is unchecked.

find-route-around-barriers/src/main/java/com/esri/arcgismaps/sample/findroutearoundbarriers/MainActivity.kt
```diff
@@ -0,0 +1,476 @@
...
+        directionsList.clear()
+
+        routeParameters?.apply {
+            // add the existing stops and barriers to the route parameters
+            setStops(stopList)
+            setPolygonBarriers(barriersList)
+        }
+
+        // solve the route task
+        val results = routeParameters?.let { routeTask?.solveRoute(it) }
```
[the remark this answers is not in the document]
prupani-7: renamed to routeResults

shubham7109 (approved):
Changes look good 👍

find-route-around-barriers/src/main/java/com/esri/arcgismaps/sample/findroutearoundbarriers/MainActivity.kt
```diff
@@ -0,0 +1,468 @@
...
+     * graphic and call showDirectionsInBottomSheet which shows directions in a list view.
+     */
+    private fun createAndDisplayRoute() = lifecycleScope.launch {
+        if (stopList.size <= 1) return@launch
+
+        // clear the previous route from the graphics overlay, if it exists
+        routeOverlay.graphics.clear()
+        // clear the directions list from the directions list view, if they exist
+        directionsList.clear()
+
```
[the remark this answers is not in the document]
prupani-7: Sounds good.

find-route-around-barriers/src/main/java/com/esri/arcgismaps/sample/findroutearoundbarriers/MainActivity.kt
```diff
@@ -0,0 +1,468 @@
...
+                onItemClickListener =
+                    AdapterView.OnItemClickListener { _, _, position, _ ->
+                        // remove any graphics that are not the original (blue) route graphic
+                        if (routeOverlay.graphics.size > 1) {
+                            routeOverlay.graphics.removeAt(routeOverlay.graphics.size - 1)
+                        }
+                        // set the viewpoint to the selected maneuver
+                        val geometry = directionsList[position].geometry
+                        geometry?.let { Viewpoint(it.extent, 20.0) }
+                            ?.let { mapView.setViewpoint(it) }
```
fpearson1111, line 408: I think it's better to just use one `let` here because the result of the first `let` will never be null, only if the geometry is null.
```suggestion
                        geometry?.let { mapView.setViewpoint(Viewpoint(it.extent, 20.0)) }
```

find-route-around-barriers/src/main/java/com/esri/arcgismaps/sample/findroutearoundbarriers/MainActivity.kt
```diff
@@ -316,21 +313,24 @@ class MainActivity : AppCompatActivity() {
...
             // add the existing stops and barriers to the route parameters
             setStops(stopList)
             setPolygonBarriers(barriersList)
         }
 
         // solve the route task
-        val routeResults = routeParameters?.let { routeTask?.solveRoute(it) }
+        val routeResults = routeParameters.let { routeTask.solveRoute(it) }
 
         if (routeResults != null) {
```
fpearson1111, line 335: Is this check still needed?

fpearson1111 (commented):
Sorry for the long delay in completing the first round of this review. I've gone through the full code now and it looks good except just a few comments.

Although I also tried the sample and for me the directions list isn't displaying anything. Can you try it out again on your device?

prupani-7:
Thanks @fpearson1111 no worries!
yes I am able to see the directions list on my device
image
I tap on the map twice, that creates 2 points and displays a route between them and the directions too

fpearson1111 (approved):
I think the reason I wasn't seeing any directions is that my phone is in dark mode. When I switch it to light mode the instructions appear. I attached some screenshots:



I think this sample should be updated so that the instructions still display in dark mode but I will approve anyway because if the fix is easy then I am happy for it to be merged without re-approval. Nice work :)

prupani-7:
I think that is something, we will need to update on the samples, which display directions list. There are 2 more samples, which will need an update I think, which displays the directions text. I will create a follow up issue, because I am not sure, how much the scope of work will be at this moment. 

The three samples which will need an update will be 
`find route around barriers`
`navigate route`
`find-route`

prupani-7:
Thanks for the feedback @fpearson1111 always helpful!

find-route-around-barriers/src/main/res/values/styles.xml
```diff
@@ -0,0 +1,11 @@
+<resources>
+
+    <!-- Base application theme. -->
+    <style name="AppTheme" parent="Theme.MaterialComponents.DayNight.DarkActionBar" >
```
shubham7109, line 4:
```suggestion
    <style name="AppTheme" parent="Theme.AppCompat.Light.DarkActionBar" >
```
For now, samples should be set to be in light theme, we can take on supporting dark theme in a future issue.

shubham7109, line 4:
```suggestion
    <style name="AppTheme" parent="Theme.MaterialComponents.Light.DarkActionBar" >
```
For now, samples should be set to be in light theme, we can take on supporting dark theme in a future issue.

prupani-7: Thanks @shubham7109 makes sense!

PR #79, opened 2023-03-13
Replace DateTime with Instant

shubham7109:
## Description
PR to replace and remove all references of Kotlin DateTime and replace them with Java Instant. 

## Links and Data
Sample Epic: `runtime/kotlin/issues/2063`

## What To Review
- Check if the use of `Instant.now()` would be the right implementation. Or `Instant.EPOCH`


sorenoid (approved):
LGTM

PR #86, opened 2023-03-24
Display Composable MapView

shubham7109:
## Description
PR to add a new Kotlin sample _"Display Composable MapView"_ in `Maps and Scenes` category.

## Links and Data
Sample Epic: `runtime/kotlin/issues/2079`


hud10837 (commented):
A couple questions

shubham7109:
Thanks @hud10837 for the helpful suggestions here. Your example app helped me rethink this sample and I've added the feedback to the sample.

The sample now switches between two viewpoints to showcase the `update` method during a remembered state change.

display-composable-mapview/src/main/java/com/esri/arcgismaps/sample/displaycomposablemapview/MapViewWithCompose.kt
```diff
@@ -0,0 +1,52 @@
...
+) {
+    AndroidView(
+        // modifiers are used to set layout parameters
+        modifier = Modifier.fillMaxSize(),
+        // the factory parameter provides a context to create a classic Android view
+        // called when the composable is created, but not when it's recomposed
+        factory = { context ->
+            MapView(context).also { mapView ->
+                // add the MapView to the lifecycle observer
+                lifecycle.addObserver(mapView)
```
hud10837, line 34: Compose actually exposes a lifecycleOwner with `LocalLifecycleOwner.current`. I briefly tested it with this and I think it works the same
```suggestion
// outside of the `AndroidView` constructor:
val lifecycleOwner = LocalLifecycleOwner.current
// inside of factory block
lifecycleOwner.lifecycle.addObserver(mapView)
```
This would remove the need to pass in a lifecycle to `MapViewWithCompose`

display-composable-mapview/src/main/java/com/esri/arcgismaps/sample/displaycomposablemapview/MapViewWithCompose.kt
```diff
@@ -0,0 +1,52 @@
...
+                    }
+                }
+            }
+        },
+
+        // update block runs every time this view is recomposed which only occurs
+        // when a `State<T>` or `MutableState<T>` parameter is changed.
+        update = { view -> // view is automatically cast to a MapView
+            view.setViewpoint(viewpoint) // called only if the viewpoint is changed
+        }
```
hud10837, line 50: I think if you are going to have the map as a parameter, then you should also update it here. Otherwise you are breaking the contract because a user might call this compose function again with a different map and as written, the map would not update. So I would add this line or remove the map parameter
```suggestion
        update = { view -> // view is automatically cast to a MapView
            view.setViewpoint(viewpoint) // called only if the viewpoint argument is changed
            view.map = arcGISMap // called only if the arcGISMap argument is changed
        }
```

display-composable-mapview/src/main/java/com/esri/arcgismaps/sample/displaycomposablemapview/MapViewWithCompose.kt
```diff
@@ -0,0 +1,52 @@
...
+        // the factory parameter provides a context to create a classic Android view
+        // called when the composable is created, but not when it's recomposed
+        factory = { context ->
+            MapView(context).also { mapView ->
+                // add the MapView to the lifecycle observer
+                lifecycle.addObserver(mapView)
+                // set the map
+                mapView.map = arcGISMap
+                // launch a coroutine to collect map taps
+                CoroutineScope(Dispatchers.Default).launch {
```
hud10837, line 38: I think `CoroutineScope()` will note inherit a coroutine context, and there is not one to inherit here anyway, so this coroutine will essentially be an async function. It would be better to use the lifecycle owner here as well:
```suggestion
                lifecycleOwner.lifecycleScope.launch(Dispatchers.Default) {
```

hud10837 (changes requested):
Looks good, it's nice to showcase these functions. I have some more thoughts on the lifecycle stuff

display-composable-mapview/src/main/java/com/esri/arcgismaps/sample/displaycomposablemapview/MapViewWithCompose.kt
shubham7109: My understanding is that any MapView property that is modified after initial composition would need to be reflected in the `update` block. Since the sample does not modify the map, it's not used in the `update` block. This is why `ArcGISMap` parameter's state is not stored by `remember` during initial composition. The only mutable state stored is the `ViewPoint` which is used in the `update` block. If my understanding is right here, this sample shouldn't be breaking any contract. 😅 


Do you think it would be helpful to showcase both `ArcGISMap` and `Viewpoint` as mutable states even if ArcGISMap is not updated? I feel like the user will have to take this into account for any MapView property that would be modified after initial composition like for example adding adding a new graphic to `graphicsOverlays` property.

shubham7109:
Thanks @hud10837 for the feedback. I've added the lifecycle feedback and changed `setViewpoint` to `setViewpointAnimated` for nicer transitions between viewpoints.

display-composable-mapview/src/main/java/com/esri/arcgismaps/sample/displaycomposablemapview/MapViewWithCompose.kt
hud10837: For me I think the contract is that a parameter to a compose function should be expected to change. What happens if the user _does_ use a stateful map argument that changes after the MapView is initialized? For this sample it's not that important, I just feel that if the map is a parameter then it should also be in the update block, unless it is called eg. `initialMap`, because then it's clear that the parameter is only for the first setup. For me I would either make the map a mutable state or remove the parameter and just hardcode the map that gets set in the factory bock.

shubham7109: > For me I think the contract is that a parameter to a compose function should be expected to change.

Agreed, this makes more sense 👍.  I've updated the code to use a stateful map which is accessed in `update`.

display-composable-mapview/src/main/java/com/esri/arcgismaps/sample/displaycomposablemapview/MapViewWithCompose.kt
```diff
@@ -46,7 +49,10 @@ fun MapViewWithCompose(
         // update block runs every time this view is recomposed which only occurs
         // when a `State<T>` or `MutableState<T>` parameter is changed.
         update = { view -> // view is automatically cast to a MapView
-            view.setViewpoint(viewpoint) // called only if the viewpoint is changed
+            lifecycleOwner.lifecycleScope.launch {
+                view.setViewpointAnimated(viewpoint) // called only if the viewpoint parameter is changed
+                view.map = arcGISMap // called only if the arcGISMap parameter is changes
```
hud10837, line 54: I guess this one can probably go outside the `launch` 😅

hud10837 (approved):
Thanks for making all these changes, looks great 👍

display-composable-mapview/README.md
```diff
@@ -0,0 +1,32 @@
+# Display Composable MapView
+
+Display a map using Jetpack Compose.
+
+![Image of display composable mapview](display-composable-mapview.png)
+
+## Use case
+
+Android Jetpack Compose is designed to be more efficient and performant than XML layouts. It allows you to define UI in a more declarative and composable way. Compose is also designed to work seamlessly with Kotlin. Visit the Android doc, [why adopt Compose](https://developer.android.com/jetpack/compose/why-adopt)?
```
eri9000, line 9:
```suggestion
Android Jetpack Compose is designed to be more efficient and performant than XML layouts. It allows you to define UI in a more declarative and composable way. Compose is also designed to work seamlessly with Kotlin. Visit the Android doc, [why adopt Compose?](https://developer.android.com/jetpack/compose/why-adopt) to learn more. 
```

display-composable-mapview/README.md
```diff
@@ -0,0 +1,32 @@
...
+## How to use the sample
+
+Run the sample to view the map. Pan and zoom to navigate the map.
+
+## How it works
+
+1. Create a custom `@Composable` function to wrap the `MapView`
+2. Use `AndroidView` to wrap a classic Android view in a Compose UI.
+3. Set it's `Modifier` to define the MapView layout parameters
+4. Use it's `factory` parameter to provide context and create `MapView(context)`
```
eri9000, line 20:
```suggestion
3. Set its `Modifier` to define the MapView layout parameters
4. Use its `factory` parameter to provide context and create `MapView(context)`
```

display-composable-mapview/src/main/java/com/esri/arcgismaps/sample/displaycomposablemapview/MainActivity.kt
```diff
@@ -0,0 +1,75 @@
...
+
+        setContent {
+            SampleAppTheme {
+                Column(
+                    modifier = Modifier.fillMaxSize(),
+                    horizontalAlignment = Alignment.CenterHorizontally
+                ) {
+                    // a mutable or immutable state is computed by remember to stored the
+                    // Composition during initial composition, and the stored value
+                    // is returned during recomposition/state change.
```
eri9000, line 56: can you clarify this a bit?

display-composable-mapview/src/main/java/com/esri/arcgismaps/sample/displaycomposablemapview/MapViewWithCompose.kt
```diff
@@ -0,0 +1,54 @@
...
+                        onSingleTap(it.mapPoint)
+                    }
+                }
+            }
+        },
+
+        // update block runs every time this view is recomposed which only occurs
+        // when a `State<T>` or `MutableState<T>` parameter is changed.
+        update = { view -> // view is automatically cast to a MapView
+            view.map = arcGISMap // called only if the arcGISMap parameter is changes
```
eri9000, line 48:
```suggestion
            view.map = arcGISMap // called only if the arcGISMap parameter is changed
```

eri9000 (commented):
LGTM, just a few suggestions.

display-composable-mapview/src/main/java/com/esri/arcgismaps/sample/displaycomposablemapview/MainActivity.kt
shubham7109: Yeah, looking at it now it could be clarified a bit more. Changed this to:
> A mutable/immutable state is computed by remember to store its value during initial composition, and updates the composition on the state value change.

eri9000 (approved):

shubham7109:
Thanks, @eri9000! Updated the branch with `v.next` and the build version to use the latest. Merging the PR :)

PR #97, opened 2023-04-11
Sample patches

shubham7109:
## Description

- Updated the `downloadvectortilestolocalcache` snippet to the correct package name
- Renamed Oauth protal URL to avoid a conflict with Generate offline map using android jetpack workmanager sample. Both samples have the same string name `<string name="portal_url">`

## Links and Data
Sample issue: `runtime/kotlin/issues/2358`


prupani-7:
Built, installed and tested the sample viewer on android 13 device.

- The offline map workmanager sample, does give out notifications for job started and completed.
- Download vector tiles samples, does not crash on startup. works as expected
- authenticate oauth sample, works as expected

PR #99, opened 2023-05-01
Patch for multiple file-providers

shubham7109:
## Description
The sample viewer seems to conflict with two samples that use `<provider>`.  Updated the provider authority and added replace overrides to avoid conflict. 

## Links and Data
Sample Epic: `runtime/kotlin/issues/2357 `

## What To Review
- Build the sample viewer using this branch to confirm if the samples run without errors. 
    - Edit feature attachments
    - Create mobile geodatabase


prupani-7 (approved):
LGTM. tested the sample viewer with the changes in the branch and the apps works fine as expected.

PR #101, opened 2023-05-02
Generate geodatabase border

shubham7109:
## Description
This PR adds a red border similar to `Generate offline map` sample and implemented a reset button to clear map.

## Links and Data
Sample Epic: `runtime/kotlin/issues/2373` 

## To Discuss
Is it possible to create a geodatabase replica with just one layer? It would be helpful to speed up the sync task if we were to only create a replica for the "Trees" [layer](https://services2.arcgis.com/ZQgQTuoyBrtmoGdP/ArcGIS/rest/services/Mobile_Data_Collection_WFL1/FeatureServer/0). 

I tried the following implementation, but the task fails with an error: 
```kotlin
defaultParameters.layerOptions.add(GenerateLayerOption(layerId = 0))
```
Error:
```log
Error fetching geodatabase: Service error code 500: Unable to create replica. Please check your parameters.
Details:
Error in creating runtime geodatabase.
```

### DIFF
--- a/generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
+++ b/generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
@@ -202,6 +202,7 @@ class MainActivity : AppCompatActivity() {
                     showError("Error creating geodatabase parameters")
                     return@launch
                 }.apply {
+                    // set the parameters to only create a replica of the Trees (0) layer
                     layerOptions.removeIf { layerOptions ->
                         layerOptions.layerId != 0L
                     }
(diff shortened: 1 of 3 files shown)


kaushikrw (commented):
@shubham7109  The `geodatabaseSyncTask.createDefaultGenerateGeodatabaseParameters()` will create parameters that already has all the layers included from the feature service. To include specific layers in the replica, you can filter through the list of `layerOptions` and keep the ones you need.

Something like this should work -
```
defaultParameters.layerOptions.removeIf { layerOption ->
    layerOption.layerId != 0L
}
```

shubham7109:
Thanks @kaushikrw, the suggestion worked perfectly. Seems like the job is completing on a faster rate than before. 

I've added in the feedback, added `updateDownloadArea()` func to show the border on map load and updated the sample to follow the button style [guidelines](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/pull/104).

generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
```diff
@@ -161,6 +201,10 @@ class MainActivity : AppCompatActivity() {
                     // show the error and return if the task fails
                     showError("Error creating geodatabase parameters")
                     return@launch
+                }.apply {
+                    layerOptions.removeIf { layerOptions ->
+                        layerOptions.layerId != 0L
+                    }
```
kaushikrw, line 207: I think some comments here as to what this does would be helpful.

kaushikrw (approved):
LGTM.

PR #102, opened 2023-05-02
Update samples new module script to use compose

shubham7109:
## Description
This PR aims to update the `NewModuleScript.jar` spin up new samples using Jetpack Compose. Here are the following changes made:
- MainActivityTemplate: Added `SampleApp()` composable function to be used as the base layout to build a sample  
- ComposeMapViewTemplate: Wraps a MapView inside an `AndroidView` and uses a ViewModel to define the state of the MapView. 
- MapViewModelTemplate: Controls the interaction and state of the MapView and performs the `update` required to reflect the changes in the state.  
- Added `SampleAppTheme` as part of the `samples-lib` to ensure all samples can use a shared theme
- Added `SampleTopAppBar` as part of the `samples-lib` reusable component to display sample title name in the top app bar.  

## To Discuss
- I used Google's [example app](https://github.com/google-developer-training/basic-android-kotlin-compose-training-unscramble) as a reference to help come up with this template. One hassle I didn't like much was that an `AndroidView` needs to have an `update {}` implementation for recomposition, but so does updating the state of a `ViewModel/AndroidViewModel`.  I don't like having to call `update` twice to reflect state changes in the MapView, I would like to hear thoughts on if there could be a better approach with this. 

https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/blob/93f8a729f01ce8893c855258201735a53b79c942/tools/NewModuleScript/ComposeMapViewTemplate.kt?plain=1#L55-L61

https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/blob/93f8a729f01ce8893c855258201735a53b79c942/tools/NewModuleScript/MapViewModelTemplate.kt#L48

## What to review
The script creates a basic template help get started with compose sample implementation. I would really like feedback to know if there could be any optimizations/implementations that would be helpful here, both from a ease of development perspective and simple enough for new users checking out samples. 

## How to test
- Run the script as follows:
```console
$ java -jar tools/NewModuleScript.jar
Enter Name of the sample with spaces (Eg. "Display New Map"):    
Display New Composable Map
Using repository $USER/../arcgis-maps-sdk-kotlin-samples
Sample Successfully Created!
```

- Optionally, you can use the example branch which has the script generated sample: https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/tree/compose-example/switch-basemaps

## Links and Data
Sample Epic: `runtime/kotlin/issues/2421 `


tools/NewModuleScript/MainActivityTemplate.kt
```diff
@@ -14,48 +14,53 @@
...
+            content = {
+                Column() {
+                    SampleTopAppBar(getString(R.string.app_name))
+                    // composable function that wraps the MapView
+                    ComposeMapView(
+                        modifier = Modifier.fillMaxSize(),
+                        mapViewModel = MapViewModel(application)
+                    )
+                }
+            })
```
sorenoid, line 64: Two suggestions: 
1) express this as a trailing lambda instead of explicitly setting `content = `. 
2) consider moving this code to a screens package and putting it in a composable called `MapScreen` or `MainScreen`. There is only one screen in this template, but it makes sense to lay out the file structure as if it could grow into a larger app without significant reorganization. The part that is missing is navigation. I think it is ok to omit navigation support since there is one screen, it would be better to implement it anyway. Short of that, I think it is ok to organize the screens so that navigation can be easily brought in.

```suggestion        
             modifier = Modifier.fillMaxSize() {
                MapScreen()
            }
```

tools/NewModuleScript/ComposeMapViewTemplate.kt
```diff
@@ -0,0 +1,62 @@
...
+import kotlinx.coroutines.launch
+/**
+ * Wraps the MapView in a Composable function.
+ */
+@Composable
+fun ComposeMapView(
+    modifier: Modifier = Modifier,
+    mapViewModel: MapViewModel,
+) {
+    val mapViewState by mapViewModel.mapViewState.collectAsState()
```
sorenoid, line 36: Consider hoisting `mapViewState` to the higher level, whether in the MapScreen I suggested earlier or to the composable where both the top app bar and map are defined. As the Map is the MapView's model, it is likely the top app bar may need to listen to it.

samples-lib/build.gradle
```diff
@@ -30,4 +34,12 @@ dependencies {
     implementation "androidx.constraintlayout:constraintlayout:$constraintLayoutVersion"
     implementation "com.google.android.material:material:$materialVersion"
     implementation "commons-io:commons-io:2.11.0"
+
+    // lib dependencies for samples using Jetpack Compose
+    implementation "androidx.core:core-ktx:$ktxAndroidCore"
+    implementation "androidx.activity:activity-compose:$composeActivityVersion"
+    implementation "androidx.compose.ui:ui:$composeUI"
+    implementation "androidx.compose.material:material:$composeMaterial"
```
sorenoid, line 42: Consider using material3. There are some differences in how things work between material and material3, and material is already out of date. It is more likely as time goes on that a user will be porting a sample to material3, and using it here can cut out a lot of friction for them.

sorenoid: consider using [the compose BOM](https://developer.android.com/jetpack/compose/bom) to specify a version for all compose dependencies.

tools/NewModuleScript/MainActivityTemplate.kt
```diff
@@ -14,48 +14,53 @@
...
-        Snackbar.make(view, message, Snackbar.LENGTH_SHORT).show()
+    @Composable
+    private fun SampleApp() {
+        Surface(
+            color = MaterialTheme.colors.background,
+            modifier = Modifier.fillMaxSize(),
+            content = {
+                Column() {
+                    SampleTopAppBar(getString(R.string.app_name))
+                    // composable function that wraps the MapView
```
sorenoid, line 58: Did you consider using a Scaffold here? It is intended for usage in this spot in the compose hierarchy.

sorenoid (changes requested):
Good stuff @shubham7109! Some thoughts

samples-lib/build.gradle
shubham7109: We plan to tackle this in the a future issue planned for samples. There are some conflicts with just switching the gradle to use Material3 and would be best to resolve them separately before updating the script.

shubham7109:
@sorenoid I've applied the PR feedback changes. The script will be helpful to get started with spinning up samples to lay out the theme guidelines using Material3. 

The script uses the `build.gradle` used in the sample "Display Composable MapView", so when updating the theme for that sample to use Material3, should update the theme used by the script as well.

tools/NewModuleScript/ComposeMapViewTemplate.kt
```diff
@@ -0,0 +1,68 @@
...
+    // get an instance of the current lifecycle owner
+    val lifecycleOwner = LocalLifecycleOwner.current
+
+    AndroidView(
+        modifier = modifier,
+        factory = { context ->
+            MapView(context).also { mapView ->
+                // add the MapView to the lifecycle observer
+                lifecycleOwner.lifecycle.addObserver(mapView)
```
sorenoid, line 49: does this observer ever get removed? You might consider adding it in a [DisposableEffect](https://developer.android.com/jetpack/compose/side-effects#disposableeffect) (I only learned about it today). It would mean moving this lifecycle initialization out of the factory block, which means the MapView would be needed before the factory block, and would need to be remembered.

sorenoid (changes requested):
Just a couple other things @shubham7109

samples-lib/build.gradle
shubham7109: Not necessarily, the theming issue is the next planned issue for the compose effort for samples and shouldn't have much affect on the script used in this PR. 

Updating to use Material3 would require changes to our `sampleslib/theme/Theme.kt` file, which would be helpful to resolve as part of the theme issue. The gradle implementation for Material3 is specific for only composable samples: `androidx.compose.material3:material3:1.1.0`. And the gradle implementation for traditional Android UI samples currently is: `com.google.android.material:material:1.7.0`. It seems like the distinction in these two material libraries shouldn't cause changes to update all the samples, just the samples using compose. 

I'm getting dependency resolution errors with with just adding material3 to the project, which tells me that there needs to be an update with the versioning used for our libraries:
```console
Duplicate class kotlin.collections.jdk8.CollectionsJDK8Kt found in modules jetified-kotlin-stdlib-1.8.10 (org.jetbrains.kotlin:kotlin-stdlib:1.8.10) and jetified-kotlin-stdlib-jdk8-1.7.20 (org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.7.20)
```

shubham7109:
@sorenoid Applied suggestion feedback and updated the JAR file to the changes. Ready for review.

sorenoid (approved):
LGTM

tools/NewModuleScript/ComposeMapViewTemplate.kt
```diff
@@ -0,0 +1,86 @@
...
+    AndroidView(
+        modifier = modifier,
+        factory = { _ ->
+            mapView.also { mapView ->
+                // launch a coroutine to collect map taps
+                lifecycleOwner.lifecycleScope.launch {
+                    mapView.onSingleTapConfirmed.collect {
+                        onSingleTap(it)
+                    }
+                }
```
kaushikrw, line 57: I think it would be better to launch such long running jobs that are tied to this composable using a `LaunchedEffect`. If the composable leaves or ends, the coroutines started within its scope will also be cancelled.

kaushikrw (changes requested):
@shubham7109 This looks great! Just some minor comments.

shubham7109:
@kaushikrw Thanks for the feedback. I have applied the changes here and updated the JAR artifact.

kaushikrw (approved):
LGTM 👍

PR #104, opened 2023-05-04
Samples buttons style

shubham7109:
## Description
This PR aims to standardize the use of buttons in sample layouts. There are quite a few quality enchancements, here is a summary of the changes: 

| Notes | New button implementation | Previous button implementation | 
|-|-|-|
| Few samples overlay a button over the `MapView`, replaced this pattern with a bottom contraint button. Samples which hide the reset button has been replaced in favor of disabled reset button for better user experience.<br/>- Create convex hull around points<br/>- Create planar and geodetic buffers<br/>- Apply function to raster from service<br/>- Clip geometry<br/>- Cut geometry<br/>- Download vector tiles<br/>- Generate geodatabase replica<br/>- Generate offline map<br/>- Generate offline map with Jetpack<br/> | image | image |
| All types of buttons were missing the "ripple" animation. Turns out setting setting the `backgroundTint` for `OutlineButton` and `RegularButton` overrides the ripple animation. Android theming [guidelines](https://m2.material.io/develop/android/theming/color) recommend using global attribute names rather than modifying the color for that component (for samples it's  the custom buttons).<br/>- The Button theme component [page](https://m2.material.io/components/buttons#theming), specifies that container color is defined by `colorPrimary` and text color is defined using `colorOnPrimary`. Modified sample's theme to use this approach, and ripple works!<br/>- Removed custom button background tint and text color xml files.<br/>- Added `IconButton` custom style for buttons with only icons.<br/>- Added `ToggleButton` custom style for toggle buttons. | new | old | 

- To resolve inconsistencies in padding/margins of buttons in samples, I've removed them in favor of a common sample-lib padding & margins. 
- Added `dimens.xml` for common button dimensions 
- Added `colorOnPrimary` for button text color and `colorOnSecondary` for FAB tint color to common App theme.  

## Guidelines used
- Samples with one button (eg. Display dimensions): Contained button constraint below the `MapView`
- Samples with two buttons (eg. Cut geometry): Contained color button for primary action and outline button for secondary action, constraint below the `MapView`. Both buttons must use up full width equally / use match constraints / `android:layout_width="0dp"`
- Avoid hiding buttons away from view, instead display a disabled button which toggles to be enabled. 
- Button text size: `15sp`
- Buttons by default should be `android:textAllCaps="false"`
- Buttons should use common app theme colors and should not set background or text colors

## Links and Data
Sample Epic: `runtime/kotlin/issues/2370 `

## What To Review
I would suggest building the sample viewer using the `v.next` build against this PR to visually compare the button changes. If there are usages that do not follow the set guidelines or could use some enhancements, feel free to provide a comment. The samples has been tested with Dark mode.


prupani-7:
Hi @shubham7109 , I started reviewing the samples and I had a concern regarding the button style we using for some samples. 

For the samples below(screenshot), I see we are using Default `contained button`  and an `outlined button`  and trying to toggle them programatically. 
Curious why we are not choosing to use the Toggle button class provided by the material library "MaterialToggleGroup"? as opposed to the way above?

Because doing it the former way, would lead to adding extra code to handle the toggle in the MainActivity.kt to enable and disable the respective buttons on click.

image

shubham7109:
@prupani-7 Thanks for taking a look at this, I have few thoughts and questions regarding your feedback. 

I see `MaterialToggleGroup` to be a component which would be useful to "toggle" between states and the selected toggle button will be in an "active" (currently selected option) state. For samples which have a primary button (eg. "Take map offline") & secondary button (eg. "Reset") I don't see them as two options with interchangeable states, instead I feel like there could be more than 2 states, for example disabling secondary button. 

Also, previously the secondary button is usually hidden away from view by programmatically modifying the `visibility` attribute. The proposed changes will instead modify the `isEnabled` attribute. I don't see any way how you could implement this using `MaterialToggleGroup` without also adding code to modify the `isEnabled` attribute. So in terms of additional code added, I cannot see a benefit with using `MaterialToggleGroup`

The proposed changes **will add** any additional code if the sample does not have a `Reset` button (eg. Generate offline map & work manager). But if the sample has been hiding away a button, then no additional Kotlin code is required aside from just modifying the `isEnabled` attribute instead of `visibility` attribute. (eg. Cut geometry & Clip geometry)

I'm assuming you feel like the proposed changes visually look similar to the `MaterialToggleGroup`, but functionally the toggle buttons are meant for a very specific use case. The Material design [doc](https://m2.material.io/components/buttons#toggle-button) for this component specifies the usage:
> Toggle buttons can be used to group related options. To emphasize groups of related toggle buttons, a group should share a common container.

This usage makes sense for samples like "Find route around barriers" to choose to add stops/barriers or "Change viewpoint" which changes between different viewpoint modes is where `MaterialToggleGroup` could be used. I'm curious to know how you would approach this to work with the affected samples listed above?

For reference, 
I used this Material Doc example as an outline to define the guidelines above. Which was to use a filled button on the right, with an outline button on it's left to help differentiate the emphasis between the two: 
image

create-planar-and-geodetic-buffers/src/main/res/layout/activity_main.xml
```diff
@@ -27,36 +27,36 @@
...
+                app:layout_constraintHorizontal_bias="0.5"
+                app:layout_constraintStart_toEndOf="@+id/clearButton"
                 app:layout_constraintTop_toTopOf="parent" />
 
-            <TextView
-                android:id="@+id/unitsDisplay"
-                android:layout_width="wrap_content"
+            <com.google.android.material.button.MaterialButton
+                android:id="@+id/clearButton"
+                style="?attr/materialButtonOutlinedStyle"
```
prupani-7, line 43:
```suggestion
                [style="?attr/materialButtonOutlinedStyle"](android:enabled="false")
```

create-planar-and-geodetic-buffers/src/main/res/layout/activity_main.xml
```diff
@@ -27,36 +27,36 @@
...
+                android:layout_width="0dp"
```
prupani-7, line 44: We can have the clearButton disabled by default and enable it when the map is tapped.
`android:enabled="false"`

create-planar-and-geodetic-buffers/src/main/java/com/esri/arcgismaps/sample/createplanarandgeodeticbuffers/MainActivity.kt
```diff
@@ -107,48 +111,65 @@ class MainActivity : AppCompatActivity() {
...
+                val geodesicBufferGraphic = Graphic(bufferGeometryGeodesic)
+
+                // create a graphic for the user tap location
+                val locationGraphic = Graphic(mapPoint)
+
+                // add the buffer polygons and tap location graphics to the appropriate graphic overlays
+                planarGraphicsOverlay.graphics.add(planarBufferGraphic)
+                geodesicGraphicsOverlay.graphics.add(geodesicBufferGraphic)
+                tapLocationsOverlay.graphics.add(locationGraphic)
             }
```
prupani-7, line 145:
```suggestion
           clearButton.isEnabled = true
           }
```

create-planar-and-geodetic-buffers/src/main/java/com/esri/arcgismaps/sample/createplanarandgeodeticbuffers/MainActivity.kt
```diff
@@ -107,48 +111,65 @@ class MainActivity : AppCompatActivity() {
...
+                bufferInMiles = value
+            }
+
+            // create and display the dialog
+            dialogBuilder.create().show()
+        }
+
         // clear the graphics from the graphics overlays
         clearButton.setOnClickListener {
             planarGraphicsOverlay.graphics.clear()
```
prupani-7, line 175:
```suggestion
            clearButton.isEnabled = false
            planarGraphicsOverlay.graphics.clear()
```

create-planar-and-geodetic-buffers/src/main/res/layout/activity_main.xml
prupani-7, line 44:
```suggestion
                android:enabled="false"
                android:layout_width="0dp"
```

download-vector-tiles-to-local-cache/src/main/res/layout/activity_main.xml
```diff
@@ -34,24 +34,21 @@
 
             <com.google.android.material.button.MaterialButton
                 android:id="@+id/exportVectorTilesButton"
-                android:layout_width="match_parent"
+                style="@style/OutlineButton"
```
prupani-7, line 37: could we also have the "Download Vector Tiles" button as a contained (regular) button and not outlined button?
Just to follow a standard set (Filled color button for primary action and outline button for secondary action) with other samples like Create planar and geodetic buffers, Apply function to raster from service, Clip geometry ..

apply-function-to-raster-from-service/src/main/res/layout/activity_main.xml
```diff
@@ -11,29 +11,46 @@
...
+                app:layout_constraintBottom_toBottomOf="parent"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintHorizontal_bias="0.5"
+                app:layout_constraintStart_toEndOf="@+id/resetButton"
+                app:layout_constraintTop_toTopOf="parent"
+                app:layout_constraintVertical_bias="1.0" />
+
+            <com.google.android.material.button.MaterialButton
+                android:id="@+id/resetButton"
+                style="@style/OutlineButton"
```
prupani-7, line 42: here we can directly use `style="?attr/materialButtonOutlinedStyle` instead of the custom style `@style/OutlineButton` as in the samples-lib/styles.xml file, we are referencing the former to use the custom "outlinebutton" style. 
See Line 73: `<item name="materialButtonOutlinedStyle">@style/OutlineButton</item>`

```suggestion
                style="?attr/materialButtonOutlinedStyle"
```

clip-geometry/src/main/res/layout/activity_main.xml
```diff
@@ -1,41 +1,53 @@
...
+                android:text="@string/clip_geometry"
+                app:layout_constraintBottom_toBottomOf="parent"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintHorizontal_bias="0.5"
+                app:layout_constraintStart_toEndOf="@+id/resetButton"
+                app:layout_constraintTop_toTopOf="parent" />
+
+            <com.google.android.material.button.MaterialButton
+                android:id="@+id/resetButton"
+                style="@style/OutlineButton"
```
prupani-7, line 39: here we can directly use `style="?attr/materialButtonOutlinedStyle` instead of the custom style `@style/OutlineButton` as in the samples-lib/styles.xml file, we are referencing the former to use the custom "outlinebutton" style. 
See Line 73: `<item name="materialButtonOutlinedStyle">@style/OutlineButton</item>`

```suggestion
                style="?attr/materialButtonOutlinedStyle"
```

similar suggestion at all other places where the MaterialButton uses `style="@style/OutlineButton"`

find-route/src/main/res/layout/activity_main.xml
```diff
@@ -38,10 +38,12 @@
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 android:layout_margin="16dp"
+                android:layout_marginEnd="30dp"
```
prupani-7, line 41: Setting `layout_margin` overrides values set for layout_marginEnd and layout_marginBottom. 

I think we can remove these.

play-kml-tour/src/main/res/layout/activity_main.xml
```diff
@@ -66,10 +67,10 @@
 
             <com.google.android.material.button.MaterialButton
                 android:id="@+id/playPauseButton"
-                android:layout_width="wrap_content"
+                android:layout_width="0dp"
                 android:layout_height="wrap_content"
-                android:text="@string/play"
                 android:enabled="false"
+                android:text="@string/play"
```
prupani-7, line 73: The `Play` button text doesn't look like its in center. I think in design, it shows its aligned centrally, but still it looks off weirdly. 

Could we make this text as "Play tour" which actually would clearly mean what we mean by "play" in context here like the "Reset tour" button does while additionally serve the purpose of the center alignment. As the "Play tour" text looks more centrally aligned.

show-grid/src/main/res/layout/activity_main.xml
```diff
@@ -1,27 +1,31 @@
...
-            android:layout_height="match_parent" />
+            android:layout_height="0dp"
+            app:layout_constraintBottom_toTopOf="@id/menuButton"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent" />
 
         <com.google.android.material.button.MaterialButton
             android:id="@+id/menuButton"
+            style="@style/OutlineButton"
```
prupani-7, line 22: Question: why did we choose this to be an Outline button and not a regular button?
It seems like a primary/main button, as the intension of this sample is to show different grids( having different label color and positions) which happens on click of this button.

prupani-7 (commented):
@shubham7109 Great work on this PR. It brings in the consistency and sets standardized rules to use Material buttons in the samples. Well done! I wonder if these guidelines we set here in this PR, need to be documented somewhere to follow them in the future samples we add to the repo. That would be really helpful imo. 

I did leave some suggestions and questions for you to look at. Let me know if we want to go over any of these on a call, happy to do so.

prupani-7:
>MaterialToggleGroup to be a component which would be useful to "toggle" between states and the selected toggle button will be in an "active" (currently selected option) state.

thanks for explaining in detail, I do see a point of using toggle group only for specific purposes, like to group related options which share a common container. Also makes sense as to why its not exactly suitable for this use case( like clip geometry etc)
 I am happy with how it is designed currently. Thank you!

show-grid/src/main/res/layout/activity_main.xml
shubham7109: The guideline on emphasis was irrelevant to me when it came to just having one button. It makes sense for a button to be "emphasized" for a layout with two or more buttons. If a layout consists of only one button, the style used could just be arbitrary, seen as in the user has no other UI interaction. 

My choice for a outlined button with singular button layout was pretty much arbitrary, although I felt it helped bring down the (for lack of a better word) purpleness of the layout in light mode. 

Either option would still respect the material theme guidelines, so it's up to us to decide. Would you prefer keeping it as an outline button or choose to make contained buttons as the default of singular button layouts? 

image

image

shubham7109:
Thanks @prupani-7 for the review. I've applied the feedback changes. Also, yes I plan to document these guidelines into a Wiki once this PR is good to be approved. Thanks for letting me know.

prupani-7 (approved):
@shubham7109 The changes look good to me. 
There is one more file (generate-geodatabase-replica-from-feature-service/src/main/res/layout/activity_main.xml) which needs an update to use the Material button style `style="?attr/materialButtonOutlinedStyle"` instead of `style="@style/OutlineButton"`

show-grid/src/main/res/layout/activity_main.xml
shubham7109: Sounds good, I've removed the outline from the following samples:
- Display dimensions
- Download vector tiles
- Show Grid 

I've updated the guideline to follow this suggestion as well. 👍

shubham7109:
@prupani-7 I've added the feedback from the comments. Also, noticed two outline buttons in the Navigate route sample, for which I just made changes to follow the guidelines. If it the additional changes looks good to you, I'll move ahead with the PR.

Thanks again for the reviews here.

prupani-7:
@shubham7109 Looks good to me 👍

fpearson1111 (approved):
LGTM. I scanned over the code and tried out a few different samples to get an idea of the changes. Everything looked ok to me (I tested only in dark mode), and I like the idea of making the sample layouts consistent. I noticed one or two things in the code but happy to approve.

PR #106, opened 2023-05-15
Search with Geocode Bug

prupani-7:
## Description

PR to fix the search with geocode bug.

## Issue
There seems to be a race condition with the search view populating more views than the result size so we end up seeing an empty view at the bottom as seen in the video below. The issue could be with the adapter not updating/refreshing the view as per the search result size

https://devtopia.esri.com/storage/user/3117/files/85658edb-4f35-4907-8afe-fb108fd53b43

## Current implementation
1. OnQueryTextChange
- Create the simple adapter cursor
- Add new row (for each result in the result.size) using the above Cursor
- Create an instance of the adapter 

## fix that works
1.  OnQueryTextChange
- Create the simple adapter cursor
- Create an instance of the adapter 
- Add new row (for each result in the result.size) using the above Cursor
- notify data change on the adapter

Branch with using the custom adapter workflow which essentially does the same logic as the above fix:
priyanka/searchwithgeocodebug

Another approach that I tried but failed: 
Is to initialize the Adapter only once on every onCreateOptionsMenu call.. (that will addRow for each result when OnQueryTextChange occurs). But if we initialize the custom adapter only once, then all the results get appended in the searchView one after other as the text is entered on the searchBar. So the adapter needs to be created every time OnQueryTextChange happens. 

## Links and Data

Sample Epic: `runtime/kotlin/issues/2372`


shubham7109 (approved):
There doesn't seem to be an optimal way to reset or clear a `MatrixCursor`. 

@prupani-7 I feel like this PR implementation good for our purposes. 👍

PR #110, opened 2023-06-05
Composable job dialog component

shubham7109:
## Description
PR to add a new composable component to display a determinate dialog for ArcGIS Jobs. 

## Links and Data
Sample Epic: `runtime/kotlin/issues/2423`

## What To Review
- Review the samples-lib component: `JobLoadingDialog.kt`
- Review using the [test app](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/tree/shubham/job-dialog-component-test/job-progress-dialog-test) to determine if the example implementation is suitable for samples. 

## How to Test
Run the [test app](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/tree/shubham/job-dialog-component-test/job-progress-dialog-test) which implements both the simple loading dialog and the job progress dialog. You can use the test app to take a area of the map offline and preview the result or test the "Cancel job" functionality. 

## Screenshots
job progress dialog test


samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/JobLoadingDialog.kt
```diff
@@ -0,0 +1,125 @@
...
+    cancelJobRequest: (Unit) -> Unit = {},
+) {
+    if (showDialog) {
+        AlertDialog(
+            onDismissRequest = {
+                // No dismiss allowed, instead use 'Cancel job' button
+            },
+        ) {
+            Surface(
+                modifier = Modifier.fillMaxWidth(),
```
prupani-7, line 64: Just a bit concerned that the dialog occupies the full width of the screen. Looking through some basic dialog examples here, https://m3.material.io/components/dialogs/guidelines
usually it is seen that the dialogs don't extend to the max width of the screen. 
In our previous samples, the dialogs used, didn't occupy the full width.
If I remove the line 64 "modifier = Modifier.fillMaxWidth(),"
the dialog automatically adjusts to default width, and then we could awlays modify some of the paddings for text and indicator for the dialog to look fine.

would like to know your thoughts on this? :)

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/JobLoadingDialog.kt
```diff
@@ -0,0 +1,125 @@
...
+@Composable
+fun PreviewJobLoadingDialog() {
+    SampleAppTheme {
+        JobLoadingDialog(
+            title = "Job dialog loading message here",
+            showDialog = true,
+            progress = 75
+        )
+    }
+}
```
prupani-7, line 125: extra line needed

prupani-7 (commented):
@shubham7109 The progress dialog looks good to me along with the test app added. I load the portal item first and then take the map offline, to test the progress dialog correctly.

Also I didn't see any timeout error when testing with this portal.

Just added one comment here for your thoughts, but mostly the PR looks good! Nice work!

prupani-7:
Just realized, that the screenshot you added in the description above, is what I added a comment above on how the dialog could potentially look. 

Interesting, that it isn't the same dialog display for when I test. It's occupied to screen's max width.

shubham7109:
@prupani-7 Got some feedback from Rama to add a pause functionality to the job dialog. Updated the dialog component in the [test app](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/tree/shubham/job-dialog-component-test/job-progress-dialog-test) as well:
clipboard

prupani-7:
@shubham7109 Thanks, I reviewed and tested the PR, the pause and resume job functionality works as expected!!

Also, as confirmed by you the alertDialog will always occupy the maxWidth of the screen on Samsung devices. Seems like a bug with samsung devices?

prupani-7 (approved):

shubham7109:
@prupani-7, yes the dialog occupying the max width is a bug with compose support on Samsung devices:
- Issue tracker: https://issuetracker.google.com/issues/260755409

Thanks for the review, merging this in now.

PR #114, opened 2023-06-10
Analyze hotspots

shubham7109:


analyze-hotspots/src/main/java/com/esri/arcgismaps/sample/analyzehotspots/components/MapViewModel.kt
```diff
@@ -0,0 +1,171 @@
...
+
+    // error dialog status
+    val errorDialogStatus = mutableStateOf(false)
+    var errorTitle = ""
+    var errorDescription = ""
+
+    /**
+     * Displays an error dialog with [title] and optional [description]
+     */
+    fun showErrorDialog(title: String, description: String = "") {
```
prupani-7, line 155: It looks like the `showErrorDialog` fun will be used by other samples too. Should we consider moving it to common samples-lib?

prupani-7 (commented):
This is a nice sample introducing new components! Good work!!

`sampleCoroutineScope` and `MessageDialog ` implementation looks neat. 
I have added some suggestions below. 

Also, once we select the "FROM" date, I think it would be good to have the "TO" date disable all dates prior to the FROM date selected, so we don't end up selecting a back date in TO DatePicker. (which would mean the date range is invalid). 

I notice, we can still `Run Analysis` with this invalid date range, a job starts but eventually errors out with below message, which isn't the error we would expect. I don't think we should try to achieve this enhancement in this PR, if its not an easy fix, but definitely worth thinking about.
image

prupani-7:
Have you noticed this? The job loading dialog, shows progress until only 10% (never reaches 100%), jobs succeeds and adds the layer on the hotspot map.

analyze-hotspots/src/main/java/com/esri/arcgismaps/sample/analyzehotspots/components/MapViewModel.kt
shubham7109: Good point, added a `MessageDialogViewModel` to the `MessageDialog` component in the samples-lib

shubham7109:
@prupani-7 Thanks for the helpful feedback, I've made the respective changes here. Looking forward for your re-review :)

prupani-7 (approved):
Looks good to me!!

shubham7109:
Updated this sample to also reflect the comments from #116, and modified the `MapViewState` to not be a state flow w.r.t this comment: https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/pull/116#discussion_r1231896846

Thanks @prupani-7 for the review. Hopefully it's okay that I've made a few changes, so you're free to suggest any changes or could leave the approval as it is.

analyze-hotspots/src/main/java/com/esri/arcgismaps/sample/analyzehotspots/components/MapViewModel.kt
```diff
@@ -0,0 +1,164 @@
...
+            // add new layer to map
+            mapViewState.arcGISMap.operationalLayers.add(hotspotMapImageLayer)
+        }?.onFailure { throwable ->
+            messageDialogVM.showMessageDialog(throwable.message.toString(), throwable.cause.toString())
+            showJobProgressDialog.value = false
+        }
+    }
+
+    fun cancelGeoprocessingJob(coroutineScope: CoroutineScope) {
+        coroutineScope.launch {
```
hud10837, line 139: Could you use `sampleCoroutineScope` here?

analyze-hotspots/src/main/java/com/esri/arcgismaps/sample/analyzehotspots/screens/BottomSheetScreen.kt
```diff
@@ -0,0 +1,117 @@
...
+import androidx.compose.ui.unit.dp
+import com.esri.arcgismaps.sample.sampleslib.theme.SampleAppTheme
+import kotlinx.coroutines.launch
+
+/**
+ * Bottom sheet layout to select a date range
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun BottomSheetScreen(
```
hud10837, line 46: Since this already is only called within a bottom sheet, maybe it could be renamed to something like `DateRangeSelector`?

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/MessageDialog.kt
```diff
@@ -0,0 +1,110 @@
...
+
+class MessageDialogViewModel : ViewModel() {
+    // display dialog when status is true
+    var dialogStatus = mutableStateOf(false)
+        private set
+
+    var messageTitle = ""
+        private set
+
+    var messageDescription = ""
```
hud10837, line 92: Should these be mutable states as well?

hud10837 (changes requested):
Looks good, a few comments

analyze-hotspots/src/main/java/com/esri/arcgismaps/sample/analyzehotspots/screens/BottomSheetScreen.kt
```diff
@@ -0,0 +1,117 @@
...
+                    modifier = Modifier.padding(16.dp),
+                    title = { Text(text = "TO") }
+                )
+            }
+        }
+    }
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Preview(showBackground = true)
```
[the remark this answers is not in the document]
shubham7109: Not really, the first `@Preview` will preview in light mode, and the second `@Preview` will preview in dark mode.

analyze-hotspots/src/main/java/com/esri/arcgismaps/sample/analyzehotspots/screens/MainScreen.kt
```diff
@@ -0,0 +1,105 @@
...
+                    analyzeHotspotsRange = { fromDateInMillis, toDateInMillis ->
+                        if (fromDateInMillis != null && toDateInMillis != null) {
+                            if (fromDateInMillis > toDateInMillis) {
+                                mapViewModel.messageDialogVM.showMessageDialog(
+                                    title = "Invalid date range",
+                                    description = "The selected \"TO\" date cannot be before the \"FROM\" date"
+                                )
+                            } else {
+                                sampleCoroutineScope.launch {
+                                    mapViewModel.apply {
```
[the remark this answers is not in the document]
shubham7109: The apply block is helpful here to reduce the noise from using the `mapViewModel` multiple times:
```kotlin
// create and run a geoprocessing task using date range
mapViewModel.createGeoprocessingJob(
    fromDate = mapViewModel.convertMillisToString(fromDateInMillis),
    toDate = mapViewModel.convertMillisToString(toDateInMillis),
)
```

Updated to directly just call the method 👍

shubham7109:
Thanks @hud10837 for the review. I've added the changes. Ready for your re-review!

analyze-hotspots/src/main/java/com/esri/arcgismaps/sample/analyzehotspots/screens/MainScreen.kt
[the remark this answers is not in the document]
hud10837: I guess it could also be an extension method on `Long`, so you could just do `fromDate = fromDateMs.toDateFormattedString()` but I'm not even sure that's better 😄

hud10837 (approved):
LGTM 👍

PR #116, opened 2023-06-13
Show coordinates in multiple formats

shubham7109:
## Description
PR to add a new Jetpack Compose sample _"Show coordinates in multiple formats"_ in `Edit and Manage Data` category.

## Links and Data
Sample Epic: `runtime/kotlin/issues/2571`

## What To Review
- Compose files and ViewModel files

## How to Test
- Run the app to test the coordinate updates on map tapped or updating the text fields. 
- I would suggest to tap the map at a location to obtain a coordinate. You can copy this coordinate and tap on a different location. Updating the text field using your copied coordinate should update the map graphic to the previews point.


show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/components/MapViewModel.kt
```diff
@@ -0,0 +1,195 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+
+class MapViewModel(application: Application) : AndroidViewModel(application) {
+    // set the MapView mutable stateflow
+    val mapViewState = MutableStateFlow(MapViewState())
+
+    var decimalDegrees = mutableStateOf("")
+    var degreesMinutesSeconds = mutableStateOf("")
+    var utm = mutableStateOf("")
+    var usng = mutableStateOf("")
```
hud10837, line 46: I did a deep dive on this as well because we usually use `StateFlow` rather than just `State`. I couldn't find a lot of info on what Android recommends as best practice, aside from [this codelab](https://developer.android.com/codelabs/jetpack-compose-state#11). I think in an Android/Compose-only app you are right to use `State`. However, I still think best practice would be to keep the setter private and to use a delegate rather than setting. So ignoring my comment above, the suggested code would be:
```suggestion
    var decimalDegrees by mutableStateOf("")
        private set
    var degreesMinutesSeconds by mutableStateOf("")
        private set
    var utm by mutableStateOf("")
        private set
    var usng by mutableStateOf("")
        private set
```
To be clear, I think you need to use `by` because if you just set the var using `=` then changing the var will not change the value inside the `MutableState` but will instead change the reference to the state itself. Using `by`, these properties can be treated as normal `var`s.

Using the `private set` ensures that only the viewmodel can change the state, but keeps the property publicly visible to observers

Note this requires two imports and anywhere you call eg. `decimalDegrees.value` would be changed to just `decimalDegrees`

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/components/MapViewModel.kt
```diff
@@ -0,0 +1,195 @@
...
+     * Uses CoordinateFormatter to update the graphic in the map from the given [coordinateNotation]
+     * string entered by the user. Also calls corresponding method to update all the remaining
+     * [coordinateNotation] strings using the notation [notationType].
+     */
+    fun fromCoordinateNotationToPoint(notationType: NotationType, coordinateNotation: String) {
+        // ignore empty input coordinate notation strings, do not update UI
+        if (coordinateNotation.isEmpty()) return
+        val convertedPoint: Point
+
+        when (notationType) {
```
hud10837, line 136: You could probably combine these ie. `val convertedPoint: Point = when (notationType) {....}`

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/components/MapViewModel.kt
```diff
@@ -0,0 +1,195 @@
...
+
+    /**
+     * Coordinate notations supported by this sample
+     */
+    enum class NotationType {
+        DMS, DD, UTM, USNG
+    }
+
+    // error dialog status
+    val errorDialogStatus = mutableStateOf(false)
```
hud10837, line 176: As above regarding the `mutableStateOf`

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/components/MapViewModel.kt
```diff
@@ -0,0 +1,195 @@
...
+            format = LatitudeLongitudeFormat.DegreesMinutesSeconds,
+            decimalPlaces = 4
+        ) ?: return showErrorDialog()
+
+        // use CoordinateFormatter to convert to Universal Transverse Mercator, using latitudinal bands indicator
+        utm.value = CoordinateFormatter.toUtmOrNull(
+            point = newLocation,
+            utmConversionMode = UtmConversionMode.LatitudeBandIndicators,
+            addSpaces = true
+        ) ?: return showErrorDialog()
```
hud10837, line 116: This is a nitpick but currently these error dialogs will show "Invalid date format" but maybe it's better if the two types of errors have different messages like "Failure to convert to Utm Coordinates"

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/screens/CoordinatesLayout.kt
```diff
@@ -0,0 +1,137 @@
...
+import com.esri.arcgismaps.sample.showcoordinatesinmultipleformats.components.MapViewModel
+
+@Composable
+fun CoordinatesLayout(
+    modifier: Modifier = Modifier,
+    onQuerySubmit: (MapViewModel.NotationType, String) -> Unit,
+    decimalDegrees: MutableState<String>,
+    degreesMinutesSeconds: MutableState<String>,
+    utm: MutableState<String>,
+    usng: MutableState<String>
```
hud10837, line 35: I understand now, you made these public `MutableState` in the VM so they could be modified here, but I still think it is a bit of a code smell. I would normally expect that composables [cannot modify any state coming in, but would instead be passed a function](https://developer.android.com/jetpack/compose/architecture#udf) which calls down to the VM which then changes the state from inside the VM. This is what you normally see in the docs for [eg. `TextField`](https://developer.android.com/jetpack/compose/text/user-input#choose-textfield).

I'm not sure what the best way to combine simplicity and standard patterns is here. Maybe a text field composable which takes in just one string and lambda to keep things clean, something like:

```kotlin
@Composable
fun CoordinatesLayout(
    modifier: Modifier = Modifier,
    mapViewModel: MapViewModel
) {
    CoordinateField(
        notationType = MapViewModel.NotationType.DD,
        text = mapViewModel.decimalDegrees,
        onTextChanged = mapViewModel::setDecimalDegrees,
        onQuerySubmit = mapViewModel::fromCoordinateNotationToPoint
    )
...
}

@Composable
fun CoordinateField(
    modifier: Modifier = Modifier,
    notationType: MapViewModel.NotationType
    text: String = "",
    onTextChanged: (String) -> Unit = {},
    onQuerySubmit: (MapViewModel.NotationType, String) -> Unit = {}
) {
      OutlinedTextField(
          modifier = modifier,
          value = text,
          maxLines = 1,
          onValueChange = onTextChanged,
          label = { Text("Decimal Degrees") },
          keyboardOptions = keyboardOptions,
          keyboardActions = KeyboardActions(
              onSearch = {
                  onQuerySubmit(notationType, text)
                  focusManager.clearFocus()
              },
          ),
      )
}
```

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/screens/MainScreen.kt
```diff
@@ -0,0 +1,77 @@
...
+import com.esri.arcgismaps.sample.showcoordinatesinmultipleformats.components.ComposeMapView
+import com.esri.arcgismaps.sample.showcoordinatesinmultipleformats.components.MapViewModel
+
+/**
+ * Main screen layout for the sample app
+ */
+@Composable
+fun MainScreen(sampleName: String, application: Application) {
+    // create a ViewModel to handle MapView interactions
+    val mapViewModel = MapViewModel(application)
```
hud10837, line 38: I think you should use the built-in factory for this, to make sure the VM is properly created/recreated by the system
```kotlin
val mapViewModel: MapViewModel = viewModel()
```

hud10837 (changes requested):
Hi @shubham7109 , the sample looks great and I love to see a sample built in Compose. I had a few thoughts on some of the patterns here.

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/MessageDialog.kt
```diff
@@ -0,0 +1,66 @@
...
+ * Composable component to display a dialog with a message
+ */
+@Composable
+fun MessageDialog(
+    title: String,
+    description: String = "",
+    showDialog: Boolean,
+    onDismissRequest: () -> Unit
+) {
+    if (showDialog) {
```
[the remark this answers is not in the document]
shubham7109: Agreed, I've removed `showDialog` and should be handled by the caller instead. This pattern of passing the `showDialog` boolean is used by multiple dialog components: 
- sampleslib.components.JobLoadingDialog
- sampleslib.components.LoadingDialog
- sampleslib.components.MessageDialog

I've updated each of them in this PR, hopefully that would be okay.

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/MessageDialog.kt
```diff
@@ -0,0 +1,66 @@
...
+        Log.e("SampleAlertMessage", "$title: $description")
+        if (description.isNotEmpty()) {
```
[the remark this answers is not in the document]
shubham7109: The reason for checking `if (description.isNotEmpty())` is because of a weird compose behavior I noticed with AlertDialogs: 

<table>
<tr>
<td> Description </td> <td> Screenshot </td>
</tr>
<tr>
<td> 
MessageDialog with description:

```kotlin
title = { Text(title) },
text = { Text(description) },
```
</td>
<td>

Screenshot 2023-06-15 at 10 24 38 AM

</td>
</tr>
<tr>
<td> 
MessageDialog with no/empty description:

```kotlin
// notice the blank gap between the title and dismiss button
title = { Text(title) },
text = { if (description.isNotEmpty()) Text(description) },
```

</td>
<td>

Screenshot 2023-06-15 at 10 25 22 AM

</td>
</tr>

<tr>
<td> 
MessageDialog with no text attribute provided for AlertDialog:

```kotlin
// no gap between title and dismiss button
title = { Text(title) },
```

</td>
<td>

Screenshot 2023-06-15 at 10 25 42 AM

</td>
</tr>
</table>

So as a weird workaround, to be table to display a dialog without the blank gap of an empty description, I've split the compose function into two `AlertDialog`'s, one for a dialog with a description and the other for a dialog without a description. I've removed the `if` check on line 32 as the description will always be not empty:
```kotlin
text = { if (description.isNotEmpty()) Text(description) }, // before 
text = { Text(description) }, // changed
```

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/MainActivity.kt
```diff
@@ -0,0 +1,59 @@
...
+        // set compose content
+        setContent {
+            SampleAppTheme {
+                SampleApp()
+            }
+        }
+    }
+
+    @Composable
+    private fun SampleApp() {
```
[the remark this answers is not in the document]
shubham7109: I plan to update the NewModuleScript with few learnings from writing these initial compose samples. I think this would be a good enhancement to have, would you suggest the function to be renamed to use just the sample title or something like this?
- `ShowCoordinatesInMultipleFormats()`
- `ShowCoordinatesInMultipleFormatsApp()`
- `ShowCoordinatesInMultipleFormatsSample()`

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/components/MapViewModel.kt
shubham7109: This was a really helpful comment. Moving forward we should design the VM in a way that only it can change the state but keep it's property publicly visible. 👍

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/components/MapViewModel.kt
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    init {
+        // create a map that has the WGS 84 coordinate system and set this into the map
+        val basemapLayer = ArcGISTiledLayer(application.getString(R.string.basemap_url))
+        val map = ArcGISMap(Basemap(basemapLayer))
+        mapViewState.value.apply {
+            arcGISMap = map
+            graphicsOverlay.graphics.add(coordinateLocation)
+        }
```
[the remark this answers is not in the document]
shubham7109: I followed this [codelab example](https://developer.android.com/codelabs/basic-android-kotlin-compose-viewmodel-and-state#4) as a guide to set up the MapView state as a `MutableStateFlow`. To me it seems like the approach of using the `MapViewState` as a mutable state (`val mapViewState by mutableStateOf(MapViewState())`) could be a viable option. 

I made some changes, and I am curious to know if you would recommend this approach moving forward or should we stick to the code lab example of using a `MutableStateFlow`?

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/screens/CoordinatesLayout.kt
shubham7109: Really nice breaking down the `CoordinatesLayout` using `CoordinateField` into it's own Compose functions. I'm beginning to understand the concept of not modifying any state coming in from the VM, but instead passing the new state as a function. This would require us to pass the instance of the `mapViewModel` to every new compose function, which should be fine. 

One point to note, I couldn't name the setter function as `setDecimalDegrees` since it would conflict with the name used by the JVM as it would throw an error:
```console
Platform declaration clash: The following declarations have the same JVM signature (setDecimalDegrees(Ljava/lang/String;)V):
    fun `<set-decimalDegrees>`(`<set-?>`: String): Unit defined in com.esri.arcgismaps.sample.showcoordinatesinmultipleformats.components.MapViewModel
    fun setDecimalDegrees(inputString: String): Unit defined in com.esri.arcgismaps.sample.showcoordinatesinmultipleformats.components.MapViewModel
```
Instead I chose to name it `setDecimalDegreesCoordinate` and similarly for the other functions.

shubham7109:
@hud10837 Thanks for the constructive review, I always really appreciate improving samples to meet standards and expectations. With the move to use Compose I'm sure there will be a significant understanding to do. 

I've introduced a `MessageDialogViewModel` as a property of the `MapViewModel` which I'm not sure goes agains't any guidelines. To me I see the `MessageDialog` component to be heavily used by samples as an approach to log errors/display messages, and I didn't see the value of handling the dismiss and show functionality as part of each sample. Instead it would be better off if it was it's own ViewModel as part of the samples-lib component. 

I've added in the feedback you provided and ready for your re-review.

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/MessageDialog.kt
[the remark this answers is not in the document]
hud10837: I see, that makes sense, very confusing on Android's part!

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/MainActivity.kt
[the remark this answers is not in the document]
hud10837: I think the default when you create a compose app in Android Studio is to use `App` so I would suggest `ShowCoordinatesInMultipleFormatsApp()`

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/components/MapViewModel.kt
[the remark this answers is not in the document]
hud10837: I think either `MutableState` or `MutableStateFlow` is okay in this scenario. Actually I had a discussion about this with Gunther yesterday regarding toolkit, and he was leaning towards not using `State` or `MutableState` in `ViewModel`s at all. I think in this case it's okay because the sample VM will probably not be copied over wholesale into somebody's app, we don't need to worry about people potentially using classic Android views and therefore needing the UI-agnostic `StateFlow`

My previous comment was actually about the class `MapViewState` though. `MapViewState` has two public, mutable properties, and the views that observe it only refer to the single instance of `MapViewState`. So I would not expect `MapViewModel.mapViewState` to be a `MutableState` because only one instance is ever used, instead I would expect `MapViewState.arcGISMap` and `MapViewState.graphicsOverlay` to be `MutableState`s. Alternatively, I would expect `MapViewState.arcGISMap and MapViewState.graphicsOverlay` to be immutable `val`s and `MapViewModel.mapViewState` to remain a `MutableState`, so that any changes to `MapViewModel.mapViewState` would be observed.

I don't think I wrote that well so here's the two versions I'm describing:

1.
```kotlin 
// Note this can't be a data class if using delegated properties
class MapViewState {
    var arcGISMap: ArcGISMap by mutableStateOf(ArcGISMap(BasemapStyle.ArcGISNavigationNight)),
    var graphicsOverlay: GraphicsOverlay by mutableStateOf(GraphicsOverlay())
}

class MapViewModel(application: Application) : AndroidViewModel(application) {
    // set the MapView state
    val mapViewState = MapViewState()
```

2.
```kotlin
data class MapViewState(
    val arcGISMap: ArcGISMap = ArcGISMap(BasemapStyle.ArcGISNavigationNight),
    val graphicsOverlay: GraphicsOverlay = GraphicsOverlay()
)

class MapViewModel(application: Application) : AndroidViewModel(application) {
    // set the MapView state
    val mapViewState by mutableStateOf(MapViewState())
```

As it is currently written, setting `MapViewModel.mapViewState.arcGISMap = ArcGISMap(BasemapStyle.ArcGISColoredPencil)` won't do anything anything. I tested this by changing the code in `onSingleTapConfirmed` to set the map, and nothing happens, and with the debugger the update block line that changes the map never gets called. That's because neither the property `MapViewModel.mapViewState` is changing, nor does `MapViewModel.mapViewState.arcGISMap` participate in the state system, so Compose doesn't know how to observe it.

In either case 1 or 2 I think that `ComposeMapView` stays exactly as is, the only difference is what things are changing--is it the state object or is it the properties on the state object? I think the biggest practical difference is how exactly the update block of the `AndroidView` will run if only one property (let's say `arcGISMap`) changes. In option 1, the update block will only run the line that updates the map. But in option 2, I think that the entire update block will run because the whole object is updated (maybe there are some optimizations under the hood). I don't think there's a real reason to do one or the other though given how small this class is. I wrote a quick test of both using the same method above and they both trigger changes to the map.

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/screens/CoordinatesLayout.kt
```diff
@@ -0,0 +1,115 @@
...
+    val focusRequester = remember { FocusRequester() }
+    // focus manager is used to clear focus from OutlinedTextField on search
+    val focusManager = LocalFocusManager.current
+    // set the keyboard options for all field to search
+    val keyboardOptions = KeyboardOptions(
+        keyboardType = KeyboardType.Text,
+        imeAction = ImeAction.Search
+    )
+
+    OutlinedTextField(
```
hud10837, line 80: This is a nitpick, but on emulator I keep trying to press enter to search when changing the coordinates, and it puts in a new line, but I think you could set `singleLine = true` here and at least it would not add the new line :D

hud10837 (changes requested):
Thanks @shubham7109 for making those changes, sorry for the long comments, just one more issue I think with the `MapViewState`

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/screens/CoordinatesLayout.kt
shubham7109: `singleLine = true` will display the text field in a single line, but while testing I noticed that on pressing enter the text will still be added into a new line (but just not visible) and will not submit query. Instead I added a `onKeyEvent` property to the modifier to detect an enter clicked to submit the query notation and coordinate.

shubham7109:
Thanks @hud10837  for the review! Added in the `MapViewState` changes and a `onQuerySubmit` on a enter key event. Ready for your re-review :)

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/components/MapViewModel.kt
```diff
@@ -208,9 +208,9 @@ class MapViewModel(application: Application) : AndroidViewModel(application) {
...
+ * Class that represents the MapView's current state
  */
-data class MapViewState(
-    var arcGISMap: ArcGISMap = ArcGISMap(BasemapStyle.ArcGISNavigationNight),
-    var graphicsOverlay: GraphicsOverlay = GraphicsOverlay()
-)
+class MapViewState {
+    var arcGISMap: ArcGISMap by mutableStateOf(ArcGISMap(BasemapStyle.ArcGISNavigationNight))
+    var graphicsOverlay: GraphicsOverlay by mutableStateOf(GraphicsOverlay())
+}
```
hud10837, line 216: This looks good, can you also change the property on the viewmodel?

hud10837 (approved):
Thanks for making those changes. I think the property on the MapViewModel could still be changed, but no need to re-review :) Looks great

show-coordinates-in-multiple-formats/src/main/java/com/esri/arcgismaps/sample/showcoordinatesinmultipleformats/components/MapViewModel.kt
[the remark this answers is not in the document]
prupani-7: @hud10837 Thanks, this was easy to understand with above explanation and examples. +1 for option 1.

version.gradle
```diff
@@ -1,6 +1,6 @@
 ext {
     // ArcGIS Maps SDK for Kotlin version
-    arcgisVersion = '200.1.0-3842'
+    arcgisVersion = '200.2.0-3904'
```
prupani-7, line 3:
```suggestion
    arcgisVersion = '200.2.0-3917'
```

prupani-7 (approved):
LGTM! 👍

PR #122, opened 2023-06-22
Add determinate download progress indicator to samples-lib

shubham7109:
## Description
PR update the samples`DownloaderActivity.kt` to use the `ArcGISEnvironment.arcGISHttpClient.download` API. This will allow all samples that require file provisioning, to download data with a progress indicator.  

## Links and Data
Sample Epic: `runtime/kotlin/issues/1995`

## What To Review
- Changed files

## How to Test
Run a sample which requires file provisioning:
- Display Dimensions (.mmpk)
- Add feature layers (.gpkg, .geodatabase, .shp)
- Find route in transport network (.zip extracted)


samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
```diff
@@ -224,43 +233,38 @@ abstract class DownloaderActivity : AppCompatActivity() {
...
-                while (downloadCount != -1) {
-                    downloadCount = byteArrayInputStream.read(data)
+            ArcGISEnvironment.arcGISHttpClient.download(
+                url = downloadURL,
+                destinationFile = File(destinationFilePath)
+            ) { totalBytes, bytesRead ->
+                if (totalBytes != null) {
+                    // update the download percentage progress
+                    val percentage = ((100.0 * bytesRead) / totalBytes).roundToInt()
+                    Handler(Looper.getMainLooper()).post {
```
gunt0001, line 249: Interesting. This should not be necessary. I wonder if we should make the progress listener lambda a suspend function, in which case you could just call something like:
```kotlin
withContext(Dispatchers.Main) {
   progressIndicator.progress = percentage
   progressTV.text = "$percentage%"
}
```

Alternativey we could make sure that the progress listener is always invoked on the main dispatcher.

@changanxian - thoughts?

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
```diff
@@ -224,43 +233,38 @@ abstract class DownloaderActivity : AppCompatActivity() {
                 emit(LoadStatus.FailedToLoad(it))
                 return@flow
             }
+            // set up the file at the download path
+            val destinationFilePath = provisionLocation.path + File.separator + portalItem.name
```
gunt0001, line 237: There are better options to create this path:
1. Use the `File` API - `val destinationFilePath = File(provisionLocation.path, portalItem.name).toString()`
2. ``val destinationFilePath = "${provisionLocation.path}File.separator${portalItem.name}"`

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
```diff
@@ -224,43 +233,38 @@ abstract class DownloaderActivity : AppCompatActivity() {
...
-                writeOutputStream.write(byteArray)
-
+            }.onSuccess {
                 // unzip the file if it is a .zip
                 if (portalItem.name.contains(".zip")) {
+                    // set up the file at the download path
+                    val provisionFile = File(destinationFilePath)
+                    // set up the zip input stream
                     val fileInputStream = FileInputStream(destinationFilePath)
                     val zipInputStream = ZipInputStream(BufferedInputStream(fileInputStream))
```
gunt0001, line 261: I believe all this unzipping logic should be implemented in a "try with resources" style, i.e. the input streams should be closed in a `finally` block. In Kotlin you would typically do this with the extension function [Closable.use](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html).

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
shubham7109: Updated this to use your suggestion. This is required since changes to the linear progress indicator is required to run on the main thread:
```kotlin
lifecycleScope.launch(Dispatchers.Main) {
    progressIndicator.progress = percentage
    progressTV.text = "$percentage%"
}

File: samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt (hunk 3 above)
Comment (shubham7109) [reply samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt]: I've updated the input streams to:
```kotlin
FileInputStream(destinationFilePath).use { fileInputStream ->
    ZipInputStream(BufferedInputStream(fileInputStream)).use { zipInputStream ->
        ...
    }
}
```

shubham7109:
Thanks for the review @gunt0001, added the feedback. Ready for your re-review.

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
changanxian: Since the listener is updating the progress UI, it needs to run on the `main` thread.  Changing the listener be `suspend` function makes more sense to me.
@gunt0001

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
```diff
@@ -224,59 +231,56 @@ abstract class DownloaderActivity : AppCompatActivity() {
                 emit(LoadStatus.FailedToLoad(it))
                 return@flow
             }
+            // set up the file at the download path
+            val destinationFilePath = File(provisionLocation.path, portalItem.name).toString()
+            // set up the download URL
+            val downloadURL =
+                "${portalItem.portal.url}/sharing/rest/content/items/${portalItem.itemId}/data"
             // get the data of the PortalItem
-            val portalItemData = portalItem.fetchData()
```
changanxian, line 228: Did you create an issue for this? we need to investigate why the usage fails.

shubham7109: The parent issue of this PR (`kotlin/issues/1995`) was created when we noticed `fetchData()` will download all the data into memory, so it was unsuitable for large file downloads. 

Should there be an individual issue to update the `fetchData()` design?

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
gunt0001: In my mind, invoking the progress listener on the main dispatcher would be evn more convenient, as the user would always be called back on the main dispatcher, so non need at all for any thread switching in user code.

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
```diff
@@ -224,59 +231,56 @@ abstract class DownloaderActivity : AppCompatActivity() {
                 emit(LoadStatus.FailedToLoad(it))
                 return@flow
             }
+            // set up the file at the download path
+            val destinationFilePath = File(provisionLocation.path, portalItem.name).toString()
```
gunt0001, line 235: Since you require a type `File` further down in your code, you might as well hold on to a `File` instance instead of `String`:
```suggestion
            val destinationFilePath = File(provisionLocation.path, portalItem.name)
```

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
```diff
@@ -224,59 +231,56 @@ abstract class DownloaderActivity : AppCompatActivity() {
...
-            runCatching {
-                val byteArrayInputStream = ByteArrayInputStream(byteArray)
-                val data = ByteArray(1024)
-                var downloadCount: Int
-                downloadCount = byteArrayInputStream.read(data)
-                while (downloadCount != -1) {
-                    downloadCount = byteArrayInputStream.read(data)
+            ArcGISEnvironment.arcGISHttpClient.download(
+                url = downloadURL,
+                destinationFile = File(destinationFilePath)
```
gunt0001, line 242: With my comment above this could be changed to:
```suggestion
                destinationFile = destinationFilePath
```

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
```diff
@@ -224,59 +231,56 @@ abstract class DownloaderActivity : AppCompatActivity() {
...
+            ) { totalBytes, bytesRead ->
+                if (totalBytes != null) {
+                    // update the download percentage progress
+                    val percentage = ((100.0 * bytesRead) / totalBytes).roundToInt()
+                    lifecycleScope.launch(Dispatchers.Main) {
```
gunt0001, line 247: This is better 👍

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
```diff
@@ -224,59 +231,56 @@ abstract class DownloaderActivity : AppCompatActivity() {
...
+                    FileInputStream(destinationFilePath).use { fileInputStream ->
+                        ZipInputStream(BufferedInputStream(fileInputStream)).use { zipInputStream ->
+                            var zipEntry: ZipEntry? = zipInputStream.nextEntry
+                            val buffer = ByteArray(1024)
+                            while (zipEntry != null) {
+                                if (zipEntry.isDirectory) {
+                                    File(provisionLocation.path, zipEntry.name).mkdirs()
+                                } else {
+                                    val file = File(provisionLocation.path, zipEntry.name)
+                                    val fout = FileOutputStream(file)
```
gunt0001, line 267: Should also make use of `use` here:
```suggestion
                                    FileOutputStream(file).use { fileOutputStream ->
```

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
```diff
@@ -224,59 +231,56 @@ abstract class DownloaderActivity : AppCompatActivity() {
...
+                                        count = zipInputStream.read(buffer)
+                                    }
+                                    fout.close()
+                                }
+
+                                zipInputStream.closeEntry()
+                                zipEntry = zipInputStream.nextEntry
                             }
-                            fout.close()
+                            zipInputStream.close()
```
gunt0001, line 279: This call should not be necessary, since the `use { }` block will take care of it.

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
shubham7109: Yeah, it should! Also removed the `fout.close()` since `.use` should be a close the output stream.

shubham7109:
Thanks for the review @gunt0001 again. Hopefully, it looks good.

gunt0001 (approved):
@shubham7109 - looks good 👍

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/DownloaderActivity.kt
```diff
@@ -224,59 +231,53 @@ abstract class DownloaderActivity : AppCompatActivity() {
...
-                val data = ByteArray(1024)
-                var downloadCount: Int
-                downloadCount = byteArrayInputStream.read(data)
-                while (downloadCount != -1) {
-                    downloadCount = byteArrayInputStream.read(data)
+            ArcGISEnvironment.arcGISHttpClient.download(
+                url = downloadURL,
+                destinationFile = destinationFile
+            ) { totalBytes, bytesRead ->
+                if (totalBytes != null) {
```
changanxian, line 244: if `totalBytes` is null, the ` progressIndicator` will be not shown, but downloading may still happen, since the server may not provide the property.  It is suggested to show a UX Spinner since we don't know how much in progress.

changanxian: maybe you could keep the old spinner to show if the `totalBytes` is null.

changanxian:
@shubham7109 looks good, one comment.

shubham7109:
Added PR feedback, and updated with `v.next` branch. Added code to modify the spinner visibility using `Dispatchers.Main`.  @changanxian Ready for your review.

changanxian:
Looks good

changanxian (approved):

PR #129, opened 2023-08-02
Geotriggers feature click fix

shubham7109:
## Description
Added a fix to the sample to invoke the correct click listener. 

## Links and Data
Sample issue: `runtime/kotlin/issues/2822 `


saratchandrakarumuri (approved):
Sample is working as expected with this fix

PR #130, opened 2023-08-02
IPS sample fix

shubham7109:
## Description
The data source used in the sample has been modified and there are two tables in the currently used [portal item](https://www.arcgis.com/home/item.html?id=8fa941613b4b4b2b8a34ad4cdc3e4bba) (previously it was 1). The feature table used in this sample updated the name from: 
- `IPS_Positioning` -> `ips_positioning`

## Links and Data
Sample issue: `runtime/kotlin/issues/2823`

### DIFF
--- a/show-device-location-using-indoor-positioning/src/main/java/com/esri/arcgismaps/sample/showdevicelocationusingindoorpositioning/MainActivity.kt
+++ b/show-device-location-using-indoor-positioning/src/main/java/com/esri/arcgismaps/sample/showdevicelocationusingindoorpositioning/MainActivity.kt
@@ -140,15 +140,19 @@ class MainActivity : AppCompatActivity() {
     private suspend fun setUpLoadTables(featureTables: MutableList<FeatureTable>) {
         featureTables.forEach { featureTable ->
             // load each FeatureTable
-            featureTable.load().onSuccess {
-                // IPS_Positioning table needs to be present
-                if (featureTable.tableName == "ips_positioning") {
-                    setupIndoorsLocationDataSource(featureTable)
-                }
-            }.onFailure {
+            featureTable.load().onFailure {
                 showError("Error loading FeatureTable: ${it.message}")
             }
         }
+        // retrieve the loaded feature tables
+        featureTables.forEach{ featureTable ->
+            // ips_positioning table needs to be present
+            if (featureTable.tableName == "ips_positioning") {
+                 return setupIndoorsLocationDataSource(featureTable)
+            }
+        }
+        // ips_positioning table not found
+        showError("Positioning Table not found in FeatureTables")
     }
 
     /**
--- a/show-device-location-using-indoor-positioning/src/main/java/com/esri/arcgismaps/sample/showdevicelocationusingindoorpositioning/MainActivity.kt
+++ b/show-device-location-using-indoor-positioning/src/main/java/com/esri/arcgismaps/sample/showdevicelocationusingindoorpositioning/MainActivity.kt
@@ -142,10 +142,8 @@ class MainActivity : AppCompatActivity() {
             // load each FeatureTable
             featureTable.load().onSuccess {
                 // IPS_Positioning table needs to be present
-                if (featureTable.tableName == "IPS_Positioning") {
+                if (featureTable.tableName == "ips_positioning") {
                     setupIndoorsLocationDataSource(featureTable)
-                } else {
-                    showError("Positioning Table not found in FeatureTables")
                 }
             }.onFailure {
                 showError("Error loading FeatureTable: ${it.message}")


saratchandrakarumuri (commented):
The original error went away. But location is not getting displayed. I am seeing the following error after the service is loaded

Screenshot_2023-08-02-17-20-57-833_com esri sampleviewer

prupani-7:
@saratchandrakarumuri which building are you testing the above portal used in the sample?
as far as I remember, that service doesnt work for building M. This webmap was from 2016 displaying building L on redlands campus. Apparently, the layout of building L has changed quite a bit since then. 

The one I test with is good in Building M, but not for general use outside Esri.
you could test by replacing line 117-123 if the sample Mainactivity.kt with below code
```
val itemUrl =
            "https://runtimecoretest.maps.arcgis.com/home/item.html?id=41fd5a7ee7044a40ad73f7f4a79a5a25"
        val authenticator = ArcGISAuthenticationChallengeHandler {
            ArcGISAuthenticationChallengeResponse.ContinueWithCredential(
                TokenCredential.create(
                    itemUrl,
                    "qtdevteam_2",
                    "qtdevteam1234",
                    0
                ).getOrThrow()
            )
        }
        ArcGISEnvironment.authenticationManager.arcGISAuthenticationChallengeHandler = authenticator
        val map = ArcGISMap(itemUrl)
        mapView.map = map
```

show-device-location-using-indoor-positioning/src/main/java/com/esri/arcgismaps/sample/showdevicelocationusingindoorpositioning/MainActivity.kt
```diff
@@ -142,10 +142,8 @@ class MainActivity : AppCompatActivity() {
             // load each FeatureTable
             featureTable.load().onSuccess {
                 // IPS_Positioning table needs to be present
-                if (featureTable.tableName == "IPS_Positioning") {
+                if (featureTable.tableName == "ips_positioning") {
```
prupani-7, line 145: umm, this is not true for the building M dataset(I used for testing purpose) that works with the sample without any code change

show-device-location-using-indoor-positioning/src/main/java/com/esri/arcgismaps/sample/showdevicelocationusingindoorpositioning/MainActivity.kt
```diff
@@ -142,10 +142,8 @@ class MainActivity : AppCompatActivity() {
...
                     setupIndoorsLocationDataSource(featureTable)
-                } else {
-                    showError("Positioning Table not found in FeatureTables")
```
prupani-7, line 148: its good to let the user know about this error, in case there is no feature table found with the above name

show-device-location-using-indoor-positioning/src/main/java/com/esri/arcgismaps/sample/showdevicelocationusingindoorpositioning/MainActivity.kt
prupani-7, line 145: The table name works for Building M service I used to test
https://runtimecoretest.maps.arcgis.com/home/item.html?id=41fd5a7ee7044a40ad73f7f4a79a5a25

saratchandrakarumuri:
@prupani-7 But if there are no IPS beacons or since I am not in Building L, it should fall back to regular location display and show the location based on GPS right. But I am not seeing any location at all... Just the building layers load and that is it

prupani-7:
@saratchandrakarumuri It does show GNSS, if you keep it running for long. I think it showed GNSS after 10mins or so
image

[The service used ](https://www.arcgis.com/home/item.html?id=8fa941613b4b4b2b8a34ad4cdc3e4bba) is the one in original sample and the table name is updated,
```
if (featureTable.tableName == "ips_positioning") {
                    setupIndoorsLocationDataSource(featureTable)
                } else {
                    showError("Positioning Table not found in FeatureTables")
                }
```

show-device-location-using-indoor-positioning/src/main/java/com/esri/arcgismaps/sample/showdevicelocationusingindoorpositioning/MainActivity.kt
prupani-7: even after I get the above error, and keep the app running for some time ~10 mins
I see the GNSS signals being detected.
Spooky behavior I say!

shubham7109:
@saratchandrakarumuri 

I can get the GPS fallback using an emulator on API level `34` consistently using either the public or internal webmap and modifying the table name to be `IPS_Positioning`. The GPS fallback on a lower API level takes longer from my testing, but I would like to see if it is a similar behavior on your end as well. 

IMAGE (3)

saratchandrakarumuri:
@shubham7109 I see similar behavior... On emulator, it is falling back to GNSS right away. But on my Android 10 device, it is not happening even after like 20 min. @prupani-7 Confirmed that on higher api level devices it is happening faster

shubham7109:
@saratchandrakarumuri Seems like an API issue on both `200.1` and `200.2` for devices running below API level 30 (including your Android 10 device). I see the same behavior in the IPS team's internal app. 

Let's treat this behavior as a separate issue since it's not relevant to the PR or the sample here.

prupani-7 (approved):
@shubham7109 The table name change looks good to me. 

although not sure why we are removing this error: showError("Positioning Table not found in FeatureTables")?

saratchandrakarumuri (approved):
Original issue is resolved with this fix and the other is being handled separately. So this is good to go

shubham7109:
@prupani-7 
> although not sure why we are removing this error: showError("Positioning Table not found in FeatureTables")?

The [portal item](https://www.arcgis.com/home/item.html?id=8fa941613b4b4b2b8a34ad4cdc3e4bba) used contains two feature tables, `ips_positioning` & `ips_recordings`. In the sample, we iterate through each table for a table matching `ips_positioning` to be able to start the indoor data source. I removed it since previously we just expect one table, but I've gone ahead and added a [commit](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/pull/130/commits/56f119bee1d32e0a71153c4bbf331a855f7b6a5b) to display the error if the table is not found.

prupani-7:
> @prupani-7
> 
> > although not sure why we are removing this error: showError("Positioning Table not found in FeatureTables")?
> 
> The [portal item](https://www.arcgis.com/home/item.html?id=8fa941613b4b4b2b8a34ad4cdc3e4bba) used contains two feature tables, `ips_positioning` & `ips_recordings`. In the sample, we iterate through each table for a table matching `ips_positioning` to be able to start the indoor data source. I removed it since previously we just expect one table, but I've gone ahead and added a [commit](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/pull/130/commits/56f119bee1d32e0a71153c4bbf331a855f7b6a5b) to display the error if the table is not found.

Thanks! 
shouldn't this check happen when featureTable.load() is success?

prupani-7:
> > The [portal item](https://www.arcgis.com/home/item.html?id=8fa941613b4b4b2b8a34ad4cdc3e4bba) used contains two feature tables, `ips_positioning` & `ips_recordings`. In the sample, we iterate through each table for a table matching `ips_positioning` to be able to start the indoor data source. I removed it since previously we just expect one table, but I've gone ahead and added a [commit](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/pull/130/commits/56f119bee1d32e0a71153c4bbf331a855f7b6a5b) to display the error if the table is not found.
> 
> Thanks! shouldn't this check happen when featureTable.load() is success?

I just saw, we are checking featureTable.load() above!

Thanks Shubham, this is good to go!

PR #136, opened 2023-09-18
Display points using clustering feature reduction

prupani-7:
## Description

PR to add a new Kotlin sample _"Display points using clustering feature reduction"_ in `Visualization` category.

## Links and Data

Sample Epic: `runtime/kotlin/issues/2926`

## What To Review

- This sample uses a "LoadingDialog" samples-lib composable to indicate that the map is loading.

- After many implementation trails to show a suitable PopUp dialog(by using composable Dialog(), samples-lib MessageDialog()), I finally resided to creating a new samples-lib component named "HtmlMessageDialog" which uses an AlertDialog to display messages with embedded HTML tags.

- To display HTML tags in a Text composable component, a helper function is created which converts the Spanned object returned from HtmlCompat.fromHtml() to an AnnotatedString. This string can then be displayed using the Text composable.
There is another way to display HTML tags, which will create a custom composable component which uses an [AndroidView with a TextView inside](https://stackoverflow.com/questions/66494838/android-compose-how-to-use-html-tags-in-a-text-view). I found the first option more convenient as it allows to directly use the Jetpack `Text` composable to display the text with html tags.

- The Readme and metadata json are different as it requires only one screenshot reference in the metadata json file for which the sample viewer can use as the thumbnail. the README_style_checker will have that difference.

## How to Test


shubham7109 (changes requested):
Nice sample @prupani-7 using the exclusive functionality introduced in v200! The sample works as expected, but I do have some thoughts on how we could design this sample best for mobile platforms. 

One helpful tip while coming up with a sample design for samples with no previous implementation, is to check out other platform's implementations and it's design workflow.  Specifically mobile implementation design ([Swift](https://github.com/Esri/arcgis-maps-sdk-swift-samples/tree/main/Shared/Samples/Display%20points%20using%20clustering%20feature%20reduction), [MAUI](https://github.com/Esri/arcgis-maps-sdk-dotnet-samples/tree/main/src/WPF/WPF.Viewer/Samples/Layers/DisplayPointsUsingClusteringFeatureReduction)).

Summary of changes requested: 
- Let's use a similar workflow for popups using a bottom sheet as designed for the Runtime SDK
- Display the popup title and the list of popup content elements
-  The sample does not seem like it is able to format `<span>` HTML content, could this be an Android compose limitation? [Google Issue](https://issuetracker.google.com/issues/135556699)

prupani-7:
@shubham7109 Thanks for the feedback and suggestions! 

Popup dialog was implemented in this sample to show each popup detail which follows the [common sample UI design](https://devtopia.esri.com/runtime/common-samples/blob/main/designs/DisplayPointsUsingClusteringFeatureReduction/README.md) which uses a `PopupManager`.

I understand Kotlin's  [PopupView issue](https://devtopia.esri.com/runtime/kotlin/issues/353) is slated for future release . Until then we can "mimic" the popup by using the material component Dialog. This way we are still following the common design sample for Maps SDK.

> One helpful tip while coming up with a sample design for samples with no previous implementation, is to check out other platform's implementations and it's design workflow. Specifically mobile implementation design

My 2 cents on this:

1. If the Kotlin team is the first to implement a particular sample design, does this mean we expect the swift and .net teams to check the implemented Kotlin sample UI design and follow the same for their platform? Are we making sure this is followed, like having cross team code reviews?

For this sample design, I referred the [common sample design](https://devtopia.esri.com/runtime/common-samples/blob/main/designs/DisplayPointsUsingClusteringFeatureReduction/README.md), and came up with the PopupDialog implementation to follow the common design UI sample. 

I also referred to the [Kotlin samples contribution guide](https://devtopia.esri.com/runtime/common-samples/wiki/Kotlin-Samples-Repo%3A-A-Contribution-Guide#Design-principles)
and see that there are 2 points mentioned about the UI design that need to be followed by developers while designing a sample
1. The sample is primarily designed to demonstrate ArcGIS Maps SDK code and while we should follow Android & Kotlin's best practices where possible, we shouldn't dedicate lots of lines of code to fancy UI or bomb-proof MVC architectural elements.
2. Design samples with the look, feel, and responsiveness of a modern Android application. Following Google's material design [standards](https://m2.material.io/design) and utilizing the material [components]
(https://m2.material.io/components?platform=android) for the sample is encouraged!


> * Let's use a similar workflow for popups using a bottom sheet as designed for the Runtime SDK

I see only iOS has this implementation. 

.Net uses a PopupView toolkit component for its mobile implementation.

display-points-using-clustering-feature-reduction/src/main/java/com/esri/arcgismaps/sample/displaypointsusingclusteringfeaturereduction/screens/MainScreen.kt
```diff
@@ -0,0 +1,106 @@
...
+                    )
+                }
+
+                // display a HtmlMessageDialog to show popup details
+                if (mapViewModel.showPopupDetailsDialog.value) {
+                    HtmlMessageDialog(
+                        description = mapViewModel.annotatedPopupString.value,
+                        onDismissRequest = { mapViewModel.showPopupDetailsDialog.value = false }
+                    )
+                }
```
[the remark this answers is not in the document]
prupani-7: > As of 200.2, the `PopUpView` is planned for a future release epic, until then the `samples-lib` could introduce a `BottomSheet` which can display a custom composable layout.

If this is the case, I would be more inclined to keep the current sample UI implementation of showing the popup details in a popup view (replicated by using a android material3 dialog) and when PopupView is available, we should create an issue to update this to use the component.

shubham7109:
> If the Kotlin team is the first to implement a particular sample design, does this mean we expect the swift and .net teams to check the implemented Kotlin sample UI design and follow the same for their platform? Are we making sure this is followed, like having cross-team code reviews?

@prupani-7 To an extent this is true! If Kotlin were to design a sample on `common-samples/design` then, every platform implementation would be based on the Kotlin sample's design. That being said, there is no strict expectation to keep the design consistent on other platforms. We have the freedom to tailor the UI suitable for our respective platforms, as long as the Maps SDK implementation remains the same as the steps detailed in the "How it works" section regardless of the platform used. The reason for my tip on referencing other platforms is so that we do not spend effort into creating a design from scratch and in my experience it does greatly help with the velocity for the sample implementation.

I also appreciate you using the Kotlin contribution guide and following the common-design implementation for this sample. I would like to point out that the sample was designed to use the Toolkit component, and has been implemented by QT (common-design), Swift, and dotNet. The suggestion to use a bottom sheet for this sample was to simply introduce a sharable component dedicated to samples similar to the design of Runtime Android PopupView, as we don't have one currently. 

I fully understand that this would be a helpful discussion to have before the sample PR process, and moving forward I would like us to take the time for such discussions. I did however provide a plug-and-place solution to avoid any overhead from your end. For this sample, feel free to choose an implementation approach that would have the best value for your time and effort.

prupani-7:
Thanks @shubham7109 for the discussion yesterday and your above comment
After our meeting, I had a chat with @rchintapalli, and we discussed that to some extent both the implementations ( Dialog vs Bottomsheet) are fine for this sample. Although he made a good point that, with Bottom sheets, once it appears on screen, it does gives the user ability to interact with the map, pan around, zoom in/out etc but with Dialog's that isn't the case and it does hide most part of the map. 
Considering these points, it does make sense to use bottom sheets to show popup details vs showing it in a Dialog.

@rchintapalli although re: this statement
"we should still dismiss the bottomsheet as soon as we start map interaction"

I think doing this, it will still make it similar to how dialog functions because when you click outside of the dialog, it dismisses and gives ability to map/zoom the map. But to you point that "Bottomsheet makes map central and more visible most of the times" still holds true.

I will go ahead and make the changes to use Bottomsheet. 
Agree that moving forward we should set the sample design expectations before we start its implementation to avoid any re-work and to avoid design related discussions in PR's

prupani-7:
@shubham7109 I used the bottom-sheet code suggestion from this PR and plugged it in my branch, and got the bottomsheet working for this sample. Thank you! Could you please review?

Although I do want to highlight that, contrary to what we thought that the mapView would still be interact-able with the bottom-sheet open, is not the case. It still behaves like a Dialog. Its advantage over Dialog is that it doesn't fully hide the map and is more visible. fyi, @rchintapalli

shubham7109:
@prupani-7 We could use `AnimatedVisibility` to animate a bottom sheet to be visible, and `mapView.onPan` & `mapView.onSingleTapConfirmed` to detect when the bottom sheet would need to be hidden. Let's discuss this over a call on we would go about implementing this on the sample:

```kotlin
@Composable
fun BottomSheet(
    bottomSheetContent: @Composable() () -> Unit,
    isVisible: Boolean
) {
    BoxWithConstraints(
        modifier = Modifier.fillMaxSize()
    ) {
        AnimatedVisibility(
            modifier = Modifier.align(Alignment.BottomCenter),
            visible = isVisible,
            enter = slideInVertically{height -> height} + fadeIn(),
            exit = slideOutVertically{height -> height} + fadeOut()
        ) {
            bottomSheetContent()
        }
    }
}
```
And on the `MainScreen` we could add the bottom sheet at the end of the `Scaffold.content`:

```kotlin
BottomSheet(
    isVisible = mapViewModel.showClusterSummaryBottomSheet.value,
    bottomSheetContent = {
        ClusterInfoContent(
            popupTitle = mapViewModel.popupTitle.value,
            clusterInfoList = mapViewModel.clusterInfoList
        ) { // on dismiss
            mapViewModel.showClusterSummaryBottomSheet.value = false
        }
    })
```

prupani-7:
@shubham7109 as discussed on our call with Rama, the above solution to dismiss the Bottomsheet on map interaction and keep the Bottomsheet open on consecutive `single-taps` of feature layers looks good with the use of `AnimatedVisibility` composable. As agreed, you could go ahead and add your implementation, in this PR branch.

shubham7109 (approved):
@prupani-7 I've updated the sample to use the `AnimatedVisibility` sheet and feel free to provide any feedback. I will approve this PR now, but one final change would be to update the screenshots for the readme file.

display-points-using-clustering-feature-reduction/src/main/java/com/esri/arcgismaps/sample/displaypointsusingclusteringfeaturereduction/components/MapViewModel.kt
```diff
@@ -0,0 +1,232 @@
...
+                        showClusterSummaryBottomSheet.value = true
+                        popup.evaluateExpressions().onSuccess {
+                            popup.evaluatedElements.forEach { popupElement ->
+                                when (popupElement) {
+                                    is FieldsPopupElement -> {
+                                        popupElement.fields.forEach { popupField ->
+                                            // convert popupField.label embedded with html tags using HtmlCompat.fromHtml
+                                            clusterInfoList.add(
+                                                HtmlCompat.fromHtml(
+                                                    popupField.label,
```
XuemingRocks, line 139: Usually, for a field in a field popup element popup shows both the label and value of the field instead of just the label. e.g. `population: 11,100`.

display-points-using-clustering-feature-reduction/src/main/java/com/esri/arcgismaps/sample/displaypointsusingclusteringfeaturereduction/screens/MainScreen.kt
```diff
@@ -0,0 +1,117 @@
...
+                    )
+                    Switch(
+                        checked = mapViewModel.isFeatureReductionEnabled.value,
+                        onCheckedChange = {
+                            mapViewModel.toggleFeatureReduction()
+                        })
+                }
+
+                // display a MessageDialog if the sample encounters an error
+                mapViewModel.messageDialogVM.apply {
```
XuemingRocks, line 90: I think this should be handled in the view model instead of in an UI element.

display-points-using-clustering-feature-reduction/src/main/java/com/esri/arcgismaps/sample/displaypointsusingclusteringfeaturereduction/screens/MainScreen.kt
```diff
@@ -0,0 +1,117 @@
...
+                }
+            }
+
+            // display a bottom sheet to show popup details
+            BottomSheet(isVisible = mapViewModel.showClusterSummaryBottomSheet.value) {
+                ClusterInfoContent(
+                    popupTitle = mapViewModel.popupTitle.value,
+                    clusterInfoList = mapViewModel.clusterInfoList
+                ){
+                    mapViewModel.showClusterSummaryBottomSheet.value = false
```
XuemingRocks, line 112: I think it should be changed by viewModel.

XuemingRocks (commented):
@prupani-7 Nice sample. Overall looks good. Just a few suggestions and a question.

display-points-using-clustering-feature-reduction/src/main/java/com/esri/arcgismaps/sample/displaypointsusingclusteringfeaturereduction/components/MapViewModel.kt
```diff
@@ -0,0 +1,232 @@
...
+        sampleCoroutineScope.launch {
+            result.onSuccess { identifyResultList ->
+                // initialize the string for each tap event resulting in a new identifyResultList
+                clusterInfoList.clear()
+                popupTitle.value = ""
+                identifyResultList.forEach { identifyLayerResult ->
+                    val popups = identifyLayerResult.popups
+                    popups.forEach { popup ->
+                        // set the popup title
+                        popupTitle.value = popup.title
```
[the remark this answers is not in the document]
prupani-7: I tested with the sample map and the map we have in example app, the popup title always remain the same which is "Cluster Summary", so think we don't need to append it each time

display-points-using-clustering-feature-reduction/src/main/java/com/esri/arcgismaps/sample/displaypointsusingclusteringfeaturereduction/screens/MainScreen.kt
prupani-7: I have followed this pattern to show MessageDialog from the MainScreen, from the other samples
CC: @shubham7109

prupani-7:
Thanks @XuemingRocks for your feedback. Applied your suggestions. Please re-review

display-points-using-clustering-feature-reduction/src/main/java/com/esri/arcgismaps/sample/displaypointsusingclusteringfeaturereduction/components/ComposeMapView.kt
```diff
@@ -0,0 +1,93 @@
...
+                setViewpoint(mapViewState.viewpoint)
+            }
+        }
+    )
+
+    // launch coroutine functions in the composition's CoroutineContext
+    LaunchedEffect(Unit) {
+        launch {
+            mapView.onSingleTapConfirmed.collect {
+                mapViewModel.showClusterSummaryBottomSheet.value = false
```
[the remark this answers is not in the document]
prupani-7: Nice catch, updated

display-points-using-clustering-feature-reduction/src/main/java/com/esri/arcgismaps/sample/displaypointsusingclusteringfeaturereduction/components/MapViewModel.kt
prupani-7: Done, updated

display-points-using-clustering-feature-reduction/src/main/java/com/esri/arcgismaps/sample/displaypointsusingclusteringfeaturereduction/screens/MainScreen.kt
XuemingRocks: I will leave this to your judgement then

display-points-using-clustering-feature-reduction/src/main/java/com/esri/arcgismaps/sample/displaypointsusingclusteringfeaturereduction/components/MapViewModel.kt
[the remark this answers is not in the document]
XuemingRocks: That is because of the popup configuration of this webmap. Some webmaps may use attribute values or expression to define popup title. But I think it is fine for this sample.

XuemingRocks (approved):
LGTM!

prupani-7:
Thanks @shubham7109 and @XuemingRocks for your reviews! Merging..

PR #137, opened 2023-09-25
Show Callout Sample

prupani-7:
## Description

PR to add a new Kotlin sample _"Show Callout"_ in `Maps` category.

## Links and Data

Sample Epic: `runtime/kotlin/issues/1117`

## What To Review

## How to Test


show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/ComposeMapView.kt
```diff
@@ -0,0 +1,105 @@
...
+
+    // launch coroutine functions in the composition's CoroutineContext
+    LaunchedEffect(Unit) {
+        launch {
+            mapView.onSingleTapConfirmed.collect {
+                // get map point from the Single tap event
+                it.mapPoint?.let { mapPoint ->
+                    // convert the point to WGS84 for obtaining lat/lon format
+                    val wgs84Point =
+                        GeometryEngine.projectOrNull(mapPoint, SpatialReference.wgs84()) as Point
```
shubham7109, line 75: I'm not a fan of Android Studio's line break formatting when it comes to lengthy code lines. My preference would be to follow the Compose formatting and add named arguments in individual lines for readability: 

```suggestion
                    val wgs84Point = GeometryEngine.projectOrNull(
                        geometry = mapPoint,
                        spatialReference = SpatialReference.wgs84()
                    ) as Point
```

shubham7109 (changes requested):
Nice sample @prupani-7, just a few minor comments mainly to use the `MapViewModel` to update and show the callout on the MapView

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/ComposeMapView.kt
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Wraps the MapView in a Composable function.
+ */
+@Composable
+fun ComposeMapView(
+    modifier: Modifier = Modifier,
+    mapViewModel: MapViewModel,
+    application: Application
+) {
```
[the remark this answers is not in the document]
prupani-7: Yeah, aware that the business logic goes in MapViewModel, and I tried that approach first, i.e. to add a new property `callout` to the MapViewState. But we cannot initialize [Callout](https://developers.arcgis.com/kotlin/api-reference/arcgis-maps-kotlin/com.arcgismaps.mapping.view/-callout/index.html?query=class%20Callout), as it has no constructors 

Thats the reason why I adapted the above implemented approach.

prupani-7:
@shubham7109 as Discussed in a call, the logic is now in mapViewModel with `calloutContent` and the `latlonpoint` properties of MapViewState and calling callout.show() in the update block of mapView.
Please re-review.

shubham7109 (approved):
Using the `MapViewModel` here looks nice! Just two minor comments, but good to approve 👍

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/ComposeMapView.kt
```diff
@@ -0,0 +1,103 @@
...
+    )
+
+    // launch coroutine functions in the composition's CoroutineContext
+    LaunchedEffect(Unit) {
+        launch {
+            mapView.onSingleTapConfirmed.collect {
+                onSingleTap(it)
+            }
+        }
+    }
```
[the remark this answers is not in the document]
prupani-7: > This seems to have been glossed over before since it is generated by the script.
Shouldn't `launch` be removed from the template itself?

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/MapViewModel.kt
```diff
@@ -0,0 +1,64 @@
...
+import com.esri.arcgismaps.sample.showcallout.R
+import kotlinx.coroutines.flow.MutableStateFlow
+
+class MapViewModel(private val application: Application) : AndroidViewModel(application) {
+    // set the MapView mutable stateflow
+    val mapViewState = MutableStateFlow(MapViewState(application))
+
+    fun onMapTapped(mapPoint: Point?) {
+        // get map point from the Single tap event
+        mapPoint?.let { mapPoint ->
```
changanxian, line 40: The argument `mapPoint` of Lamda closure shadows the `mapPoint`, it'd better change a different name, e.g 'point'

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/ComposeMapView.kt
```diff
@@ -0,0 +1,101 @@
...
+        // enable animated callout
+        callout.isAnimationEnabled = true
+    }
+
+    // wrap the MapView as an AndroidView
+    AndroidView(
+        modifier = modifier,
+        factory = { mapView },
+        // recomposes the MapView on changes in the MapViewState
+        update = { mapView ->
```
changanxian, line 60: same here, The argument `mapView` of Lamda closure shadows the variable mapView, it'd better change a different name.

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/ComposeMapView.kt
```diff
@@ -0,0 +1,101 @@
...
+import kotlinx.coroutines.launch
+
+/**
+ * Wraps the MapView in a Composable function.
+ */
+@Composable
+fun ComposeMapView(
+    modifier: Modifier = Modifier,
+    mapViewModel: MapViewModel,
+    onSingleTap: (SingleTapConfirmedEvent) -> Unit = {}
```
changanxian, line 41: Could it be better for the callback be a `suspend` function, ` onSingleTap: suspend (SingleTapConfirmedEvent) -> Unit = {}`

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/MapViewModel.kt
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+/**
+ * Data class that represents the MapView state
+ */
+data class MapViewState(val application: Application) {
+    var arcGISMap: ArcGISMap = ArcGISMap(BasemapStyle.ArcGISNavigationNight)
+    var viewpoint: Viewpoint = Viewpoint(34.056295, -117.195800, 1000000.0)
+    var calloutContent: TextView by mutableStateOf(TextView(application))
```
changanxian, line 62: It is a little odd for me to have a UI `TextView` in the class mix in other non-UI classes, could it be refactored into a String? Keep the UI TextView outside of the class, we only need to keep track of the state of the contents, right?

changanxian:
@prupani-7 a few comments.

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/ComposeMapView.kt
prupani-7: I am not sure if I understand, by what you mean here. By `shadows the mapView variable` do you mean the one created above `val mapView = createMapViewInstance() ` on line 48?
This is something we follow across all samples, and part of the sample template.

To me `mapView` still holds valid in the update block. 

If at all, we decide to make a change, I would vote for making change to the mapView var created using createMapViewInstance() to something like `mapViewInstance` or `mapViewObject`

CC: @shubham7109

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/ComposeMapView.kt
prupani-7: @changanxian I don't think I understand, what value will it bring in current samples, making the onSingleTap callback a suspend function?

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/MapViewModel.kt
prupani-7: Good point, this got me thinking that the 2 variables `calloutContent` and `latLonPoint` could be MapViewModel properties instead of `MapViewState` properties. 

image
To
image

TextView needs an applicationContext available in the viewmodel class

@shubham7109 Since we worked on this change together, tagging you here for review.

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/ComposeMapView.kt
changanxian: I mean the line 60, the `mapView` overshadows the outside variable `mapView`

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/ComposeMapView.kt
changanxian: It may not be necessary in this case, and keep the current implementation.  But it will be an asynchronous call. Imagine if the `onSingleTap` is called in a loop, it will be different.

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/ComposeMapView.kt
```diff
@@ -0,0 +1,101 @@
...
+                    }
+                }
+            }
+        }
+    )
+
+    // launch coroutine functions in the composition's CoroutineContext
+    LaunchedEffect(Unit) {
+        mapView.onSingleTapConfirmed.collect {
+            onSingleTap(it)
```
changanxian, line 81: After this Friday's coffee meeting(9/29), I think you need to change this line to
```
launch {
                onSingleTap(it)
            }
```

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/ComposeMapView.kt
```diff
@@ -0,0 +1,101 @@
...
+                        // center the map on the tapped location
+                        setViewpointCenter(latlonPoint)
+                    }
+                }
+            }
+        }
+    )
+
+    // launch coroutine functions in the composition's CoroutineContext
+    LaunchedEffect(Unit) {
```
changanxian, line 79: It is a little better to change this to `LaunchedEffect(mapView) `

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/ComposeMapView.kt
changanxian: It is OK for this case, `onSingleTap` doesn't block anything

show-callout/src/main/java/com/esri/arcgismaps/sample/showcallout/components/ComposeMapView.kt
changanxian: it may be not necessary.

changanxian:
@prupani-7 Looks good

prupani-7:
Thanks @changanxian @shubham7109 for your reviews. Merging...

PR #138, opened 2023-10-04
fix sample viewer build "Visualization" category crash

prupani-7:
[Sample viewer build sample category crash](https://devtopia.esri.com/runtime/kotlin/issues/2926#issuecomment-4363153):
Fixes an error with sample viewer build caused by the feature reduction metadata file, with the order of the images file names.
Ran the sample viewer app, with this change to verify, and the Visualization category doesnt crash anymore.


shubham7109 (approved):
Nice 👍  seems like we need to keep in mind to use the main screenshot as the first `image` when using multiple screenshots.

PR #165, opened 2024-02-08
Foreground service type patch - v.next

ruiqima:
## Description

PR to fix the crashing of sample `generate-offline-map-using-android-jetpack-workmanager` when running on Android 14 (API level 34) devices.

## Links and Data

#3463

## What To Review

- The correct ForegroundServiceType is used. View all available type [here](https://developer.android.com/develop/background-work/services/fg-service-types#data-sync).
- Sample `generate-offline-map-using-android-jetpack-workmanager` is running on both Android 13 and 14 devices, and the job can execute as expected.

## How to Test


shubham7109 (approved):
Looks good to approve. Sample is working fine on the latest and the old API versions for me.

puneet-pdx (approved):

PR #166, opened 2024-02-08
Foreground service type patch - feature branch

ruiqima:
## Description

PR to fix the crashing of sample `generate-offline-map-using-android-jetpack-workmanager` when running on Android 14 (API level 34) devices.

## Links and Data

#3463

## What To Review

**Same changes as in the PR for `v.next`**.

- The correct ForegroundServiceType is used. View all available type [here](https://developer.android.com/develop/background-work/services/fg-service-types#data-sync).
- Sample `generate-offline-map-using-android-jetpack-workmanager` is running on both Android 13 and 14 devices, and the job can execute as expected.

## How to Test


shubham7109:
You can close this PR. This sample is not using compose, so it doesn't need to be updated in the compose feature branch. 

All the compose samples from this branch will eventually be merged to `v.next` which will already have the foreground service patch for this sample.

ruiqima:
Close this PR since we don't need to update it in the compose feature branch.

PR #172, opened 2024-03-05
Merge v.next into feature branch to use kotlin DSL

ruiqima:
## Description

PR to merge v.next changes into feature branch. Mainly update to use Kotlin DSL.

## Links and Data

#3647

## What To Review

Some samples may fail to run, check if those samples are recorded in the Revisit Sample epic. 

## How to Test

Run each samples to check.


shubham7109 (approved):
Tested this by running this branch with the sample viewer, and the samples that fail are as follows:

- add-dynamic-entity-layer
- analyze-hotspots
- identify-layer-features
- manage-operational-layers
- query-feature-table
- show-coordinates-in-multiple-formats
- show-magnifier

These are planned to be resolved individually. Approving this PR 👍

version.gradle
```diff
@@ -1,34 +0,0 @@
-ext {
```
prupani-7, line 1: This file got deleted in the feature branch. Also the build.gradle files from the affected samples in the PR.
@ruiqima

PR #194, opened 2024-04-03
[New Sample] Snap Geometry Edits

Luiy0:
## Description

This PR is for the `Snap Geometry Edits` sample in the `Edit and Manage Data` category. 

This sample design aims to showcase feature snapping functionality while editing a geometry using the `GeometryEditor`.

## Links and Data
Design: [SnapGeometryEdits](https://devtopia.esri.com/runtime/common-samples/pull/3900)
Data: [WaterDistributionNetwork](https://www.arcgis.com/home/item.html?id=b95fe18073bc4f7788f0375af2bb445e)

## What To Review
- Readme, metadata, and resource files.
- Review the code to make sure it is easy to follow like other samples on Android.

## How to Test
- Run the sample prototype on the sample viewer or the repo.

https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/assets/72321786/33317a7a-a90b-4768-acdc-9ba344087bed


snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/screens/MainScreen.kt
```diff
@@ -0,0 +1,192 @@
...
+                    mapViewProxy = mapViewModel.mapViewProxy,
+                    mapViewInteractionOptions = MapViewInteractionOptions(isMagnifierEnabled = true),
+                    onSingleTapConfirmed = mapViewModel::identify,
+                    onPan = { mapViewModel.dismissBottomSheet() }
+                )
+                Row(
+                    modifier = Modifier
+                        .padding(12.dp)
+                        .fillMaxWidth(),
+                ) {
```
shubham7109, line 97: This `Row` can be extrapolated into a Composable function outside of the `MainScreen` function.  Maybe something along the lines of: `GeometryOptionsMenu(mapViewModel: MapViewModel)` which consumes the ViewModel.

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/components/MapViewModel.kt
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    /**
+     * Creates a graphic from the geometry and add it to the GraphicsOverlay.
+     */
+    private fun createGraphic() {
+        val geometry = geometryEditor.stop()
+        val graphic = Graphic(geometry)
+
+        when (geometry!!) {
```
shubham7109, line 172: Using `!!` not-null assertion operator may not be ideal if the geometry is null. You could use an elvis operator to return an error message if `geometry` is null. 

```suggestion
        // stop the editing session and retrieve created geometry 
        val geometry = geometryEditor.stop()
            ?: return messageDialogVM.showMessageDialog("Error stopping editing session")
        // convert the geometry into a graphic type.
        val graphic = Graphic(geometry)
        
        when (geometry) {
```

shubham7109 (changes requested):
Nice sample @Luiy0 , overall looks good. Tested by building it with the sample viewer. Just few minor quality comments:

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/components/MapViewModel.kt
```diff
@@ -0,0 +1,247 @@
...
+            is Point -> graphic.symbol = GeometryEditorStyle().vertexSymbol
+            is Multipoint -> graphic.symbol = GeometryEditorStyle().vertexSymbol
```
shellygill, line 174: 🔧  You can put the two conditions on one like, like this:
```suggestion
            is Point, is Multipoint -> graphic.symbol = GeometryEditorStyle().vertexSymbol
```

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/components/MapViewModel.kt
```diff
@@ -0,0 +1,247 @@
...
+            is Polyline -> graphic.symbol = GeometryEditorStyle().lineSymbol
+            is Polygon -> graphic.symbol = GeometryEditorStyle().fillSymbol
+            is Envelope -> graphic.symbol = GeometryEditorStyle().lineSymbol
```
shellygill, line 177: Geometry from a GeometryEditor will not be an `Envelope`.

```suggestion
            else -> {}
```

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/screens/MainScreen.kt
```diff
@@ -0,0 +1,192 @@
...
+                            )
+                        }
+                    }
+                    val vector = ImageVector
+                    IconButton(
+                        enabled = mapViewModel.isUndoButtonEnabled.value,
+                        onClick = { mapViewModel.editorUndo() }
+                    ) {
+                        Icon(vector.vectorResource(R.drawable.undo), contentDescription = "Undo")
+                    }
```
shellygill, line 148: 🤡  It's typical if you have an undo button that you also have a redo button.

shellygill (commented):
Thanks @Luiy0 - a few minor comments from me, but I leave it to the Kotlin team for official approval.

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/components/MapViewModel.kt
```diff
@@ -0,0 +1,247 @@
...
+        dismissBottomSheet()
+    }
+
+    /**
+     * Enables or disables the UI buttons when an editing session is started or stopped.
+     */
+    private fun toggleButtons() {
+        isCreateButtonEnabled.value = !isCreateButtonEnabled.value
+        if (isCreateButtonEnabled.value) {
+            isUndoButtonEnabled.value = false
```
shellygill, line 238: Missed this one earlier - the Undo button should be enabled or disabled based on the value of `canUndo`.

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/screens/MainScreen.kt
Luiy0: I tried this approach and will need to look at it further because the buttons now become enabled before the layers are loaded when moving the row to an outside composable function.

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/screens/MainScreen.kt
Luiy0: The design aims to have a simplified geometry editor so that we can highlight snapping. I agree that the expectation would be to have both undo and redo but at the same time question if it's necessary to have both. We can diverge from the design and add the redo if that's preferred.

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/screens/MainScreen.kt
Luiy0: I was able solve this by ensuring the map is loaded before loading the operational layers. Thanks @fpearson1111 for the help.

Luiy0:
Hi all, thanks for the reviews. I have added the suggested changes.

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/screens/MainScreen.kt
shellygill: I'd think either having both, or else neither if simpler code is the ideal makes most sense to me, but having said that I'm fine to go with the concensus if only having undo was the preferred approach.

shellygill (commented):
Updates look OK to me @Luiy0 thanks.

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/components/MapViewModel.kt
```diff
@@ -0,0 +1,255 @@
...
+
+    // create boolean flags to track the state of UI components
+    val isUndoButtonEnabled = geometryEditor.canUndo
+    val isCreateButtonEnabled = mutableStateOf(false)
+    val isSaveButtonEnabled = mutableStateOf(false)
+    val isDeleteButtonEnabled = mutableStateOf(false)
+    val isSnapSettingsButtonEnabled = mutableStateOf(false)
+    val isBottomSheetVisible = mutableStateOf(false)
+    val snappingCheckedState = mutableStateOf(false)
+    val snapSourceCheckedState = mutableStateListOf(false)
```
fpearson1111, line 74: Hmm shouldn't this just be an empty list initially, like `mutalbeStateListOf()`, and then it gets populated below in `showBottomSheet`?

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/components/MapViewModel.kt
```diff
@@ -0,0 +1,255 @@
...
+     */
+    fun stopEditor() {
+        if (identifiedGraphic.geometry != null) {
+            identifiedGraphic.geometry = geometryEditor.stop()
+            identifiedGraphic.isSelected = false
+        } else {
+            if (geometryEditor.isStarted.value) {
+                createGraphic()
+            }
+        }
```
fpearson1111, line 168:
```suggestion
        } else if (geometryEditor.isStarted.value) {
            createGraphic()
        }
```

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/components/MapViewModel.kt
```diff
@@ -0,0 +1,255 @@
...
+                returnPopupsOnly = false
+            ).getOrNull()
+
+            if (!geometryEditor.isStarted.value) {
+                if (graphicsResult != null) {
+                    if (graphicsResult.isNotEmpty()) {
+                        identifiedGraphic = graphicsResult[0].graphics[0]
+                        identifiedGraphic.isSelected = true
+                        identifiedGraphic.geometry?.let { geometryEditor.start(it) }
+                        toggleButtons()
```
fpearson1111, line 233: nitpick: if this nullability check is done before starting the editor then it should also be done for toggling the buttons or else they could get out of sync
```suggestion
                        identifiedGraphic.geometry?.let {
                            geometryEditor.start(it)
                            toggleButtons()
                        }
```

fpearson1111 (commented):
Nice sample! I have a few comments but I have used the sample a lot and I know it works well so don't worry if you can't get to all of these

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/components/MapViewModel.kt
```diff
@@ -0,0 +1,255 @@
...
+
+    // create lists for displaying the snap sources in the bottom sheet
+    private val _snapSourceSettingsList = MutableStateFlow(listOf<SnapSourceSettings>())
+    val snapSourceList: StateFlow<List<SnapSourceSettings>> = _snapSourceSettingsList
+
+    // create boolean flags to track the state of UI components
+    val isUndoButtonEnabled = geometryEditor.canUndo
+    val isCreateButtonEnabled = mutableStateOf(false)
+    val isSaveButtonEnabled = mutableStateOf(false)
+    val isDeleteButtonEnabled = mutableStateOf(false)
```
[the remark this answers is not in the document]
Luiy0: The createButton is a bit different in that it also depends on when the layers are loaded but the other two can be updated making the `toggleButtons()` method unnecessary.

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/components/BottomSheetContent.kt
```diff
@@ -0,0 +1,245 @@
...
+import com.arcgismaps.geometry.GeometryType
+import com.arcgismaps.mapping.layers.FeatureLayer
+import com.arcgismaps.mapping.view.geometryeditor.SnapSourceSettings
+import com.esri.arcgismaps.sample.sampleslib.theme.SampleTypography
+
+/**
+ * Composable component to display the snapping configuration settings.
+ */
+@Composable
+fun SnapSettings(
```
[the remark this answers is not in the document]
Luiy0: I'm not sure, @shubham7109 do you have any suggestions?

fpearson1111 (approved):
LGTM

shubham7109 (approved):
Looks good to me, just the one comment on the file name and package change. 

@fpearson1111 @shellygill Thanks for the helpful review here!

PR #197, opened 2024-04-22
Bug: Update searchView query with selected suggestion result

ruiqima:
## Description

PR to resolve bug that the textfield is not filled when choosing a text from suggestion list.

## Links and Data

#3259

## What To Review

Sample `search-with-geocode` and `geocode-offline` run well. When selecting from the suggestion list, the selected text is filled in textfield and can search successfully.

## How to Test

Sample `search-with-geocode` and `geocode-offline` run well.


shubham7109 (approved):
Looks good 👍 

No need to get two reviews for this, feel free to merge in.

PR #210, opened 2024-06-28
Trev8939/update o auth sample to toolkit

TADraeseke:


authenticate-with-oauth/src/main/java/com/esri/arcgismaps/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -17,103 +17,50 @@
...
+
+        Surface(
+            color = MaterialTheme.colorScheme.background
+        ) {
+            MainScreen(
+                sampleName = getString(R.string.app_name)
+            )
+            // Displays appropriate Authentication UI when an authentication challenge is issued.
+            // Because the authenticatorState has an oAuthUserConfiguration set, authentication
+            // challenges will happen via OAuth.
```
gunt0001, line 62: I suggest to add here some explanation about where in your app `DialogAuthenticator` should be called, something like:
```suggestion
            // challenges will happen via OAuth.
            // Call the DialogAuthenticator composable function at the top level of your view hierarchy, for example
            // at the same level as MainScreen(). This ensures that authentication handling is set up before any
            // components of the ArcGIS Maps SDK that may require authentication are used.
```

authenticate-with-oauth/README.md
```diff
@@ -12,71 +12,57 @@ Your app may need to access items that are only shared with authorized users. Fo
...
-5. Set the `AuthenticationManager`'s `arcGISAuthenticationChallengeHandler` to the `ArcGISAuthenticationChallengeHandler` created above.
-6. Load a map with premium content requiring authentication to automatically invoke the `ArcGISAuthenticationHandler`.
-
-#### Setting up the Manifest.xml:
-
-1. Set the launch mode for the `OAuthUserSignInActivity` to route it's intent instance through a call to its `onNewIntent()` method, rather than creating a new instance of the activity. To find out more on setting the launch configuration of an activity, visit the [Android docs](https://developer.android.com/guide/topics/manifest/activity-element)
+1. Create an `AuthenticatorState` and set the `OAuthConfiguration` with the portal URL, client ID, and redirect URI.
+2. Load a `PortalItem(...)` with connection type `Portal.Connection.Authenticated` which will issue an authentication challenge.
```
gunt0001, line 18: I would swap the order of step 2 and 3. In my mind the logical order should be that `DialogAuthenticator()` is called before loading the portal item, so that the OAuth prompt can actually happen.

authenticate-with-oauth/README.md
```diff
@@ -12,71 +12,57 @@ Your app may need to access items that are only shared with authorized users. Fo
...
+3. Create a `DialogAuthenticator` composable passing it the `AuthenticatorState` object. Because the `AuthenticatorState` object has an `OAuthConfiguration` set, the `DialogAuthenticator` will prompt for OAuth credentials.
```
gunt0001, line 19: Instead of "Create" I suggest to use the term "Call" here, since we are calling a composable function. Also would mention "toolkit":
```suggestion
3. Call the toolkit's `DialogAuthenticator` composable passing it the `AuthenticatorState` object. Because the `AuthenticatorState` object has an `OAuthConfiguration` set, the `DialogAuthenticator` will prompt for OAuth credentials when the associated `Portal` is loaded.
```

authenticate-with-oauth/README.metadata.json
```diff
@@ -14,12 +13,12 @@
         "credential",
         "portal",
         "security",
+        "toolkit",
         "ArcGISAuthenticationChallengeHandler",
         "ArcGISAuthenticationChallengeResponse",
         "AuthenticationManager",
-        "OAuthUserConfiguration",
```
gunt0001, line 20: `OAuthUserConfiguration` is still used in this sample, so I wouldn't remove it.

authenticate-with-oauth/README.md
```diff
@@ -12,71 +12,57 @@ Your app may need to access items that are only shared with authorized users. Fo
...
 ## Relevant API
 
 * ArcGISAuthenticationChallengeHandler
 * ArcGISAuthenticationChallengeResponse
 * AuthenticationManager
-* OAuthUserConfiguration
-* OAuthUserCredential
-* OAuthUserSignIn
+* AuthenticatorState
+* DialogAuthenticator
```
gunt0001, line 55: Somewhere in this doc, probably more at the top, it should be highlighted that this sample makes use of the toolkit's `authentication` module, possibly with a [link](https://github.com/Esri/arcgis-maps-sdk-kotlin-toolkit/tree/main/toolkit/authentication) to that module in the toolkit repo.

gunt0001 (commented):
@TADraeseke - looks good, some minor comments. Great simplification of the OAuth sample! 👍

TADraeseke:
Thanks @gunt0001 -- made those changes!

gunt0001 (approved):
@TADraeseke - Nice one 👍

eri9000 (commented):
Great work @TADraeseke 
I left some suggestions

authenticate-with-oauth/src/main/java/com/esri/arcgismaps/sample/authenticatewithoauth/MainActivity.kt
```diff
@@ -17,103 +17,54 @@
...
-            redirectUrl = getString(R.string.oauth_sample_redirect_uri)
-        )
-    }
+class MainActivity : ComponentActivity() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        // authentication with an API key or named user is
+        // required to access basemaps and other location services
+        ArcGISEnvironment.apiKey = ApiKey.create(BuildConfig.API_KEY)
```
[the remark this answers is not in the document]
TADraeseke: Good point!

PR #213, opened 2024-07-02
Use ReticleVertexTool in snapGeometryEdits

Luiy0:
## Description
Updates the `snap geometry edits` sample to use the `ReticleVertexTool` by default. 

- Use `ReticleVertexTool` by default.
- Remove magnifier.
- Fix deleting mid-vertices, reorder buttons.

Todo: 
- Any readme updates based on design.

## Links and Data
[Design PR](https://devtopia.esri.com/runtime/common-samples/pull/3932)

## What To Review
Code, readme/metadata

## Screenshots

Screenshot 2024-07-02 at 18 47 57


snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/components/MapViewModel.kt
```diff
@@ -201,7 +204,8 @@ class MapViewModel(
             isCreateButtonEnabled.value = true
         }
 
-        if (geometryEditor.selectedElement.value != null) {
+        val selectedElement = geometryEditor.selectedElement.value
+        if (selectedElement != null && selectedElement.canDelete) {
```
shellygill, line 208: 👍

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/screens/ButtonMenu.kt
```diff
@@ -107,18 +107,18 @@ fun ButtonMenu(
...
         ) {
             Icon(Icons.Filled.Delete, contentDescription = "Delete")
         }
+        IconButton(
+            enabled = mapViewModel.isSaveButtonEnabled.collectAsState().value,
+            onClick = { mapViewModel.stopEditor() }
+        ) {
+            Icon(vector.vectorResource(R.drawable.save), contentDescription = "Save")
+        }
         Row(
```
shellygill, line 122: Is there a reason why this has changed order?

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/screens/MainScreen.kt
```diff
@@ -64,7 +64,6 @@ fun MainScreen(sampleName: String) {
                     geometryEditor = mapViewModel.geometryEditor,
                     graphicsOverlays = graphicsOverlayCollection,
                     mapViewProxy = mapViewModel.mapViewProxy,
-                    mapViewInteractionOptions = MapViewInteractionOptions(isMagnifierEnabled = true),
```
shellygill, line 67: 👍

shellygill (commented):
Minor comments for you @Luiy0 .

Were you planning to hold off the merge until the common sample design is final so you can add the readme updates in this same PR too?

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/screens/ButtonMenu.kt
Luiy0: Given saving is the last thing we do when we're done editing, it made sense for it to be the right-most icon. Also, because the delete button can delete selected vertices, it can be used more often during editing.

Luiy0:
When using only the `ReticleVertexTool` we're not able delete whole geometries as the tool doesn't select them but we can delete individual vertices when these are picked up. Is this good enough for the sample or do we want to be able to delete whole geometries too which might involve adding the vertex tool?

shellygill (approved):
LGTM @Luiy0 .

> When using only the ReticleVertexTool we're not able delete whole geometries anymore as the tool doesn't select them but we can delete individual vertices when these are picked up. Is this good enough for the sample as it's mostly concerned with editing vertices using snapping rather than editing whole geometries?

Yes I think that's OK, as you say it's really the snapping of vertices we're specifically focusing on in this sample as snapping is only applicable to vertices. And you can still use the delete button to delete a point geometry vertex resulting in an empty geometry and then the delete, undo, and save buttons are disabled, which is consistent with how things behave for other geometry types.

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/components/MapViewModel.kt
```diff
@@ -81,6 +82,8 @@ class MapViewModel(
         sampleCoroutineScope.launch {
             // set the id for the graphics overlay
             graphicsOverlay.id = "Editor Graphics Overlay"
+            // set the tool for the geometry editor
```
TADraeseke, line 85:
```suggestion
            // set the tool for the geometry editor to use a reticle
```

snap-geometry-edits/src/main/java/com/esri/arcgismaps/sample/snapgeometryedits/components/MapViewModel.kt
TADraeseke, line 208:
```suggestion
       if (selectedElement?.canDelete == true) {
```

This checks null and reads easier, potentially

TADraeseke (commented):
Looks good @Luiy0! My only high level question is (and it's probably beyond the scope of your changes). Seems like snapping is the main thing we want users to see, so why isn't snapping on by default when we run this sample?

Luiy0:
>Seems like snapping is the main thing we want users to see, so why isn't snapping on by default when we run this sample?

Thanks for the review @TADraeseke! Your question came up during design discussions too and I think the idea is for users to understand how snapping can be configured by choosing to enable snap sources from the bottom sheet.

TADraeseke (approved):
LGTM!

shubham7109 (approved):
ReticleVertexTool is very fun to use, and a significant improvement from the magnifier for mobile devices. Looks good to me! 

Tested this on the sample viewer ([vTest](https://runtime-kotlin.esri.com/view/all/job/vtest/job/sampleviewer/7/)) and works as expected. 

Feel free to merge this in to `v.next`.

PR #226, opened 2024-07-25
Geodatabase replica bug

shubham7109:
## Description
PR to update the Generate geodatabase replica sample to function on reset. 

Summary of changes:
- Close the replica geodatabase before creating a new one. 
- FeatureLayer is now displayed on launch to preview features within the extent before creating a replica. 

### Links
[#4339](https://devtopia.esri.com/runtime/kotlin/issues/4339)


generate-geodatabase-replica-from-feature-service/src/main/java/com/esri/arcgismaps/sample/generategeodatabasereplicafromfeatureservice/MainActivity.kt
```diff
@@ -114,6 +129,8 @@ class MainActivity : AppCompatActivity() {
 
         // set the button's onClickListener
         generateButton.setOnClickListener {
+            // close the current geodatabase, if a replica was already generated
+            geodatabase?.close()
```
gunt0001, line 133: Note, you can only close a gdb if it's not in use anymore. Is the map that displays data of this gdb still being displayed at this point here? If so, it's not safe to call close here.

PR #237, opened 2024-10-07
[Patch] Create and save map

01smito01:
## Description

Updates Create and Save Map to unset API key & invalidate any existing OAuth credentials, making it play nice with the sample viewer.

## Links and Data

Sample Epic: [#2922](https://devtopia.esri.com/runtime/kotlin/issues/2922)
  - [x] link: [Yep](https://runtime-kotlin.esri.com/view/all/job/vtest/job/sampleviewer/45/)

## What To Review
- Is it sensible to use `runBlocking` like this? I had wanted to use `lifetimeScope`, but that caused the authenticator to loop indefinitely when re-running the sample in the viewer. This may cause a small delay when re-running the sample in the viewer.
- Does the sample work properly in the viewer?
  - In particular, does it behave as expected before and after running "Authenticate with OAuth", which sets an API key and authenticates the user?

## How to Test
Run the sample in the sample viewer.

### DIFF
--- a/create-and-save-map/src/main/java/com/esri/arcgismaps/sample/createandsavemap/MainActivity.kt
+++ b/create-and-save-map/src/main/java/com/esri/arcgismaps/sample/createandsavemap/MainActivity.kt
@@ -23,17 +23,27 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Surface
 import androidx.compose.runtime.Composable
 import androidx.lifecycle.viewmodel.compose.viewModel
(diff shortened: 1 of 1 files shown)


shubham7109 (approved):
LGTM 👍  Thanks @01smito01

shubham7109:
@01smito01 Tested using the build from the vTest, both samples load without errors. We should use a similar fix for the authenticate-with-oauth & show-portal-user-info samples, to prevent the cached oauth login to persist between samples.

PR #246, opened 2024-10-17
Fix get sample usage

shubham7109:
### Description

PR to fix the get sample usage using the sample title string for the `SearchSuggestionsList` screen


TADraeseke (approved):
LGTM!

I double checked those two get functions were unused in the project and looks like they aren't

PR #248, opened 2024-10-23
Simple fix to not re-running load

TADraeseke:
Explored dependency injection, which would require importing dagger/hilt and is probably beyond min ship.

How's this for a simple check to prevent repeat calls to load logic?

### DIFF
--- a/app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/SampleViewerApplication.kt
+++ b/app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/SampleViewerApplication.kt
@@ -1,0 +1,16 @@
+package com.esri.arcgismaps.kotlin.sampleviewer
+
+import android.app.Application
+import com.esri.arcgismaps.kotlin.sampleviewer.model.DefaultSampleInfoRepository
+import kotlinx.coroutines.runBlocking
+
+class SampleViewerApplication: Application() {
+
+    override fun onCreate() {
+        super.onCreate()
+        runBlocking {
+            // Load the repository once at app launch
+            DefaultSampleInfoRepository.load(applicationContext)
+        }
+    }
+}
--- a/app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/SampleViewerApplication.kt
+++ b/app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/SampleViewerApplication.kt
@@ -2,14 +2,15 @@ package com.esri.arcgismaps.kotlin.sampleviewer
 
 import android.app.Application
 import com.esri.arcgismaps.kotlin.sampleviewer.model.DefaultSampleInfoRepository
-import kotlinx.coroutines.MainScope
+import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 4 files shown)


shubham7109:
I would maybe recommend using this logic in the MainActivity, to be clear that `load` is only meant to run once per app session. 
```kotlin
var isInitialized by rememberSaveable { mutableStateOf(false) }
LaunchedEffect(Unit) {
    if (!isInitialized) {
        DefaultSampleInfoRepository.load(context)
        isInitialized = true
    }
}
```

shubham7109 (approved):
LGTM 👍

app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/model/DefaultSampleInfoRepository.kt
```diff
@@ -36,62 +35,61 @@ import kotlinx.serialization.json.Json
...
-        val json = Json { ignoreUnknownKeys = true }
-        context.assets.list("samples")?.forEach { samplePath ->
-            // Get this metadata files as a string
-            context.assets.open("samples/$samplePath/README.metadata.json").use { inputStream ->
-                val metadataJsonString = inputStream.bufferedReader().use { it.readText() }
-                try {
-                    val metadata = json.decodeFromString<SampleMetadata>(metadataJsonString)
+        if (isInitialized.compareAndSet(false, true)) {
+            // Iterate through the metadata folder for all metadata files
+            val json = Json { ignoreUnknownKeys = true }
```
sorenoid, line 49: This should be initialized as a property outside of the function because it is an expensive call and thus a simple performance improvement.

app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/SampleViewerApplication.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import android.app.Application
+import com.esri.arcgismaps.kotlin.sampleviewer.model.DefaultSampleInfoRepository
+import kotlinx.coroutines.runBlocking
+
+class SampleViewerApplication : Application() {
+
+    override fun onCreate() {
+        super.onCreate()
+        runBlocking {
```
sorenoid, line 11: this will block the main thread at app launch. Better to launch this, as you had it before.

sorenoid (changes requested):
LGTM, Two small suggestions to consider.

app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/model/DefaultSampleInfoRepository.kt
```diff
@@ -36,6 +36,8 @@ import kotlinx.serialization.json.Json
  */
 object DefaultSampleInfoRepository : SampleInfoRepository {
 
+    private var isInitialized = false
```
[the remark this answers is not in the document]
TADraeseke: Hmm, this was already in place. Not sure why it didn't come through

TADraeseke:
Apologies for missing your earlier comment on instantiating the json object.

Ready for re-review @sorenoid!

PR #257, opened 2024-11-07
Fix sample name inconsistencies 

shubham7109:
## Description
PR to fix the sample title not being displayed for a few select samples. This is due to the sample viewer string reserving `app_name` ([here](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/blob/v.next/app/src/main/res/values/strings.xml#L2)) causing any sample using the same string value to not display the sample title but the viewer title. 

The solution was to rename the sample title string so as not to use the reserved string name. 

## Links and Data
Issue: [#4928](https://devtopia.esri.com/runtime/kotlin/issues/4928)


samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/JobLoadingDialog.kt
shubham7109, line 1: On landscape mode this dialog does not render as expected, here is the visual changes: 

### Before: 


### After:

TADraeseke (approved):
LGTM! I opened this PR to address title sentence casing issues: https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/pull/259

sorenoid (approved):

PR #273, opened 2024-11-19
Handle SampleViewer exceptions

shubham7109:
## Description
PR to add logic to handle uncaught exceptions gracefully

Source: https://www.droidcon.com/2022/04/05/gracefully-handling-android-app-crashes/

## Links and Data
Issue: [#5017](https://devtopia.esri.com/runtime/kotlin/issues/5017)

## How to Test
- Write code to intentionally crash a sample when interacted with or launched. 

https://github.com/user-attachments/assets/6bb22329-877f-4194-97dd-02cf0d1df3ba


app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/SampleViewerLauncherActivity.kt
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     *  Prepares extras for the intent with error information if a [throwable] is provided.
+     *  It then starts the [MainActivity] by creating an intent with necessary flag and extras.
+     */
+    private fun startMainActivity(throwable: Throwable?) {
+        val extras = Bundle()
+        if (throwable != null) {
+            extras.putStringArray(
+                /* key = */     "SampleViewerException",
+                /* value = */   arrayOf(throwable.message.toString(), throwable.cause.toString())
```
TADraeseke, line 42: Do we want to separate out additionalMessage or any others? Might aid in formatting with /n's between each in the description section of the MessageDialog.

Maybe just fine as is...

TADraeseke (approved):
Really cool fail safe to have if something falls over!

app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/SampleViewerLauncherActivity.kt
shubham7109: The `Throwable` does not give us properties to access the `additionalMessage` individually but rather is concentrated into the `.cause` property. If the override param was an `Exception` it could have been possible to cast to an `ArcGISException`  to make this work.

shubham7109: Leaving it as is 👍

app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/SampleViewerLauncherActivity.kt
```diff
@@ -0,0 +1,86 @@
...
+            uncaughtException(t, e)
+        }
+    }
+}
+
+/**
+ * This interface defines a method [uncaughtException] to handle uncaught exceptions.
+ */
+interface ExceptionListener {
+    fun uncaughtException(thread: Thread, throwable: Throwable)
```
sorenoid, line 85: thread is unused, should it be removed?

app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/MainActivity.kt
```diff
@@ -22,22 +22,47 @@ import androidx.activity.compose.setContent
...
+     * Responsible for managing incoming exceptions and displaying it as a dialog in the UI.
+     */
+    @Composable
+    fun HandleExceptions() {
+        var sampleViewerException by remember { mutableStateOf(arrayOf<String>()) }
+        LaunchedEffect(Unit) {
+            intent.extras?.let {
+                sampleViewerException = it.getStringArray("SampleViewerException") ?: arrayOf()
+            }
+        }
```
sorenoid, line 59: This seems like it could be simplified. the `remember` call will run once per composition as will the LaunchedEffect so the logic in the LaunchedEffect could be run in the remember call.

```suggestion
        var sampleViewerException by remember { 
            mutableStateOf(
                intent.extras?.let {
                    it.getStringArray("SampleViewerException") ?: arrayOf()
                } ?: arrayOf<String>()
            ) 
        }
```

sorenoid (changes requested):
Nice solution, @shubham7109! I have some minor feedback.

TADraeseke:
Changes made--ready for re-review @sorenoid. Thanks!

sorenoid (approved):

PR #290, opened 2024-12-10
Update MapViewModelTemplate.kt

TADraeseke:
with TODO reminding that the mutablestate object is only needed if the map is expected to change in this sample

Came out of coffee meeting topic where it's presence in the template was leading to confusion

### DIFF
--- a/tools/NewModuleScript/MapViewModelTemplate.kt
+++ b/tools/NewModuleScript/MapViewModelTemplate.kt
@@ -28,6 +28,7 @@ import com.esri.arcgismaps.sample.sampleslib.components.MessageDialogViewModel
 import kotlinx.coroutines.launch
 
 class MapViewModel(application: Application) : AndroidViewModel(application) {
+    // TODO - The ArcGISMap only needs to be wrapped in a MutableState object if it will change in this sample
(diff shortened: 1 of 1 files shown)


01smito01 (approved):
LGTM - I'm not quite sure what the message should be, but this will do.

tools/NewModuleScript/MapViewModelTemplate.kt
```diff
@@ -28,6 +28,7 @@ import com.esri.arcgismaps.sample.sampleslib.components.MessageDialogViewModel
 import kotlinx.coroutines.launch
 
 class MapViewModel(application: Application) : AndroidViewModel(application) {
+    // TODO - The ArcGISMap only needs to be wrapped in a MutableState object if it will change in this sample
```
colinanderson, line 31: Could be
"TODO - delete mutable state when the map does not change or the screen does not need to observe changes".

colinanderson (approved):
@TADraeseke wording a message is pretty difficult. Approving as this is an improvement but I'm not convinced we should have this as the default. Surely if you create a sample that needs to observe the map changing you will work out/already know how to do that?

PR #307, opened 2025-01-16
sample complete

TADraeseke:
## Description

PR to add a new Kotlin sample "Get elevation at point on surface" in `Scenes` category.


shubham7109 (approved):
Looks good 👍

PR #324, opened 2025-02-19
Show viewshed from point on map

shubham7109:
## Description
PR to add a new Kotlin sample _"Show viewshed from point on map"_ in `Analysis` category.

## Links and Data
Sample Epic: `runtime/kotlin/issues/ISSUE_NUMBER`
  - [ ] link:

## What To Review


TADraeseke (changes requested):
Looks good! Just a few comments below @shubham7109!

shubham7109:
@TADraeseke Thanks for the review. I have applied the feedback changes, and the PR is now ready for your review!

samples/show-viewshed-from-point-on-map/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## How it works
+
+1. Create a `GeoprocessingTask` object with the URL set to a geoprocessing service endpoint.
+2. Create a `FeatureCollectionTable` object and add a new `Feature` object whose geometry is the viewshed's observer `Point`.
+3. Make a `GeoprocessingParameters` and pass in the `GeoprocessingFeatures` table which contains the observation point as an `input` parameter.
+4. Use the geoprocessing task to create a `GeoprocessingJob` object with the parameters.
+5. Start the job and wait for it to complete and return a `GeoprocessingResult` object.
+6. Get the resulting `GeoprocessingFeatures` object.
+7. Iterate through the viewshed features to use their geometry or display the geometry in a new `Graphic` object.
```
eri9000, line 23: I suggest mentioning somehow that the `GeoprocessingFeatures` are the viewshed features?
```suggestion
7. Iterate through the viewshed features (GeoprocessingFeatures) to use their geometry or display the geometry in a new `Graphic` object.
```

eri9000 (commented):
@shubham7109 Good work,
only one suggestion: 
Do you think it will be useful if we add a buffer (circle transparent buffer) showing the visibility range of the point, in this case `15km`?
This way, the user will get reminded that this is the range being taking into consideration

samples/show-viewshed-from-point-on-map/src/main/java/com/esri/arcgismaps/sample/showviewshedfrompointonmap/components/ShowViewshedFromPointOnMapViewModel.kt
eri9000, line 1: where are you defining that the viewshed is 15km?

shubham7109: Looking at the [service](https://sampleserver6.arcgisonline.com/arcgis/rest/services/Elevation/ESRI_Elevation_World/GPServer/Viewshed), it seems like we have an optional parameter `Viewshed_Distance` with a default value of 5km. We somehow missed adding this in the sample design. Nice catch 👍

shubham7109:
Thanks for the helpful comments and feedback @eri9000. I have applied the changes, ready for your review 👍

PR #332, opened 2025-03-17
Sample build JSON asset data

shubham7109:
## Description
PR to address build files source code misleading android studio. The changes in this PR aims to not use local copied files, but instead maintain a JSON sample asset data with all (100) samples assets in a `sampleAssets/samples.json`:  

```json
{
   "sampleA": {
     "MainActivity.kt": "...content...",
     "README.md": "...content...",
     ...
   },
   "sampleB": {
      ...
   }
}
```

- Note: Run a build `clean` before running the changes in this branch. 

## Links and Data
Sample Epic: [#5467](https://devtopia.esri.com/runtime/kotlin/issues/5467)
  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/sampleviewer/89/

### DIFF
--- a/app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/model/DefaultSampleInfoRepository.kt
+++ b/app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/model/DefaultSampleInfoRepository.kt
@@ -19,6 +19,7 @@ package com.esri.arcgismaps.kotlin.sampleviewer.model
 import android.content.Context
 import android.util.Log
 import com.esri.arcgismaps.kotlin.sampleviewer.model.Sample.Companion.loadActivityPath
+import com.esri.arcgismaps.kotlin.sampleviewer.model.Sample.Companion.loadCodeFiles
 import com.esri.arcgismaps.kotlin.sampleviewer.model.Sample.Companion.loadReadMe
 import com.esri.arcgismaps.kotlin.sampleviewer.model.Sample.Companion.loadScreenshot
 import com.esri.arcgismaps.kotlin.sampleviewer.model.room.AppDatabase
@@ -33,7 +34,7 @@ import kotlinx.serialization.json.Json
 import java.util.concurrent.atomic.AtomicBoolean
 
 /**
- * The single source of truth for app wide data. It reads the sample metadata to create as list of
+ * The single source of truth for app wide data. It reads the sample metadata to create a list of
  * [Sample] objects and populates the database used for search.
  * It also provides functions to get samples by category, name, or all samples.
  */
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/model/Sample.kt
```diff
@@ -62,41 +62,25 @@ data class Sample(
...
          */
-        fun loadReadMe(context: Context, sampleName: String): String {
-            // Get this metadata files as a string
-            context.assets.open("samples/$sampleName/README.md").use { inputStream ->
-                val readMeString = inputStream.bufferedReader().use { it.readText() }
-                // Remove screenshot markdown text from the README
-                return readMeString.lines().filterNot { it.contains("![") }.joinToString("\n")
-            }
+        fun loadReadMe(fileMap: Map<String, String>): String {
+            val readMeString = fileMap["README.md"] ?: return ""
```
TADraeseke, line 81: I think we want this to fail more loudly if null

app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/model/DefaultSampleInfoRepository.kt
```diff
@@ -47,51 +48,51 @@ object DefaultSampleInfoRepository : SampleInfoRepository {
...
+            val samplesJsonString = context.assets.open("samples/samples.json").use { stream ->
+                stream.bufferedReader().use { it.readText() }
+            }
+
+            // Parse it into a map of: sampleFolderName -> mapOf( filename -> fileContent )
+            val allSamplesData = json.decodeFromString<Map<String, Map<String, String>>>(samplesJsonString)
+
+            // Build each Sample using the metadata from "README.metadata.json"
+            allSamplesData.forEach { (sampleFolderName, fileMap) ->
+                val metadataJsonString = fileMap["README.metadata.json"] ?: return@forEach
```
TADraeseke, line 69: If we're getting a null for the metadata, that'd be a big enough problem we probably want to throw

app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/model/Sample.kt
```diff
@@ -62,41 +62,25 @@ data class Sample(
...
-                    ).bufferedReader().use { it.readText() }
-                    codeFiles.add(
-                        CodeFile(
-                            name = sampleAssetFile,
-                            code = codeString
-                        )
-                    )
+            fileMap.forEach { (fileName, fileContent) ->
+                if (fileName.endsWith(".kt", ignoreCase = true)) {
+                    codeFiles.add(CodeFile(name = fileName, code = fileContent))
```
TADraeseke, line 71: Nice!

gradle-plugins/src/main/kotlin/com/arcgismaps/CopySampleFilesTask.kt
```diff
@@ -42,6 +48,58 @@ class CopySampleFilesTask : Plugin<Project> {
...
+
+            // After the copy finishes, run JSON generation + cleanup
+            doLast {
+                // Create a JSON file that maps each sample folder’s files to their contents
+                createJsonAssetFile(outputDir)
+
+                // Delete everything but the JSON file itself
+                outputDir.listFiles()?.forEach { file ->
+                    if (file.isDirectory || file.name != "samples.json") {
+                        file.deleteRecursively()
```
TADraeseke, line 60: So this should handle the transition for users on the current sample viewer--their sample metadata, code files, READMEs should all get deleted.

gradle-plugins/src/main/kotlin/com/arcgismaps/CopySampleFilesTask.kt
```diff
@@ -42,6 +48,58 @@ class CopySampleFilesTask : Plugin<Project> {
 
             // Place in the assets codeFiles folder of the app.
             into(outputDir)
+
+            // After the copy finishes, run JSON generation + cleanup
+            doLast {
```
TADraeseke, line 53: I'm not great with gradle--what's the advantage of doing this part of the work at task execution time instead of configuration time

TADraeseke (changes requested):
Seems like a really good solution @shubham7109 -- just a couple comments and questions. Thanks!

gradle-plugins/src/main/kotlin/com/arcgismaps/CopySampleFilesTask.kt
shubham7109: Users (downloading from PlayStore), will not need to run the `CopySampleFilesTask`, but instead the SV will be pre-loaded with the `samples.json` in the assets folder. It would be when users download a new app-version  `Room` would detect a change in the [versionCode](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/blob/main/app/src/main/java/com/esri/arcgismaps/kotlin/sampleviewer/model/room/AppDatabase.kt#L30), which should make a new instance of the DB with the JSON file changes. Which will be responsible for handling the transition for users on the current sample viewer.

gradle-plugins/src/main/kotlin/com/arcgismaps/CopySampleFilesTask.kt
shubham7109: Placing the code for the JSON generation and cleanup in a `doLast` block ensures that it only happens if/when the `copyCodeFiles` configuration task actually runs. This avoids doing cleanup on every single Gradle invocation.

If the task is up-to-date and Gradle skips it, the code in `doLast` won't run. This is helpful because we don’t want to regenerate the JSON or delete files every time Gradle configures tasks. Gradle can quickly skip tasks that are up-to-date during execution time.

shubham7109:
Thanks for the helpful feedback @TADraeseke. Ready for your review!

gradle-plugins/src/main/kotlin/com/arcgismaps/CopySampleFilesTask.kt
TADraeseke: Awesome, sounds good. Thanks @shubham7109

gradle-plugins/src/main/kotlin/com/arcgismaps/CopySampleFilesTask.kt
TADraeseke: Yeah I mean for users who have a local copy from github, which I think is probably still the majority use case of samples.

TADraeseke (approved):

gradle-plugins/src/main/kotlin/com/arcgismaps/CopySampleFilesTask.kt
shubham7109: Tested switching `v.next` and this branch while only running a sync before a build and the updated cleanup task was able to handle the deletion process.

colinanderson (approved):
@shubham7109 I tried this out and it definitely seems to fix the problems 👍

PR #346, opened 2025-04-08
Make AR module optional for the sample viewer

shubham7109:
## Description
PR to make the AR module an optional module for the sample viewer and add logic to handle the scenario if the device does not support AR. 

## Links and Data
Sample Epic: #[5638](https://devtopia.esri.com/runtime/kotlin/issues/5638)
  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/sampleviewer/100/

### DIFF
--- a/samples/augment-reality-to-show-tabletop-scene/src/main/AndroidManifest.xml
+++ b/samples/augment-reality-to-show-tabletop-scene/src/main/AndroidManifest.xml
@@ -4,7 +4,8 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.CAMERA" />
 
-    <!-- Limits app visibility in the Google Play Store to ARCore supported devices
+    <!-- Provide support for ARCore via Google Play Services,
+         by checking to see if device is compatible and ready for AR.
          (https://developers.google.com/ar/devices). -->
     <uses-feature
         android:name="android.hardware.camera.ar"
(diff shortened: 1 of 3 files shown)


samples/augment-reality-to-show-tabletop-scene/src/main/AndroidManifest.xml
```diff
@@ -6,7 +6,9 @@
 
     <!-- Limits app visibility in the Google Play Store to ARCore supported devices
```
TADraeseke, line 7: Do we need to consider altering this comment. The line below now doesn't limit visibility in the store, but it could be a useful flag to indicate to users where/how they do limit

TADraeseke: I guess the comment on line 28/29 is still valid

TADraeseke (approved):
Looks good to me. I was blocked (without a crash) when running an API 28 emulate device and all worked well on modern device--that it redirected me to the google play store to download ar core.

samples/augment-reality-to-show-tabletop-scene/src/main/AndroidManifest.xml
shubham7109: Updated comment to say: 
```xml
    <!-- Provide support for ARCore via Google Play Services,
         by checking to see if device is compatible and ready for AR.
         (https://developers.google.com/ar/devices). -->
```

shubham7109:
@TADraeseke, it turns out the solution I originally had didn't resolve the issue on GooglePlay. I added a new commit change following Google's doc to make AR core optional: https://developers.google.com/ar/develop/java/enable-arcore#ar-optional 

Test this change on the play store, and no issues were found. Requesting a re-review.

TADraeseke (approved):
LGTM!

PR #426, opened 2025-09-24
Style point with distance composite scene symbol

shubham7109:
## Description
PR to add a new Kotlin sample _"Style point with distance composite scene symbol"_ in `Scenes` category.

## Links and Data
Sample issue: `kotlin/issues/6563`
  - [x] link: [vTest](https://runtime-kotlin.esri.com/view/vTest/job/vtest/job/sampleviewer/197/)


gunt0001 (commented):
@shubham7109 - implementation looks good, nice work. I did some ad-hoc testing and noticed that the app crashes/freezes every time I rotate the screen. Could you test this on your end.

shubham7109:
@gunt0001 Nice catch, took me some thinking to unravel the behavior: 

> noticed that the app crashes/freezes every time I rotate the screen

Turns out the `OrbitGeoElementCameraController` will trigger a crash if the geoElement the controller is orbiting is not part of the visible graphics overlay of the SceneView. On rotation, previous graphics overlay containing the plane graphic is replaced with a new one, which causes the controller to trigger a crash. 

Added some improvements to dispose of the old controller, and create a new controller for each lifecycle keeping the graphic overlay displayed up-to-date with the orbit controller provided to the SceneView.  (Using `GlobeCameraController` as the default fallback.)

shubham7109:
Updated PR to implement latest toolkit build to implement fixes. PR is ready for review.

samples/style-point-with-distance-composite-scene-symbol/src/main/java/com/esri/arcgismaps/sample/stylepointwithdistancecompositescenesymbol/screens/StylePointWithDistanceCompositeSceneSymbolScreen.kt
```diff
@@ -0,0 +1,90 @@
...
+        topBar = { SampleTopAppBar(title = sampleName) }
+    ) { padding ->
+        Column(modifier = Modifier.padding(padding)) {
+            // Provide the scene, camera controller, overlays and the SceneViewProxy.
+            SceneView(
+                modifier = Modifier.fillMaxSize().weight(1f),
+                arcGISScene = viewModel.arcGISScene,
+                sceneViewProxy = viewModel.sceneViewProxy,
+                cameraController = viewModel.orbitCameraController,
+                graphicsOverlays = listOf(viewModel.graphicsOverlay),
```
gunt0001: Setting the graphicsoverlays like this is not a good idea, since you create a new list instance with every recomposition. You could fix that by using `remember` but since we use a view model anyway, I suggest you expose the collection of graphics overlays on the view model as a property `graphicsOverlays` and then:
```suggestion
                graphicsOverlays = viewModel.graphicsOverlays,
```

gunt0001 (approved):
@shubham7109 - works great. One more comment, otherwise look good.

sorenoid (approved):

PR #439, opened 2025-10-01
Proposal: Message dialog improvement

gunt0001:
## Description

Kotlin issue: #6652


shubham7109 (commented):
I would welcome these changes as it helps standardize the type of dialog to show, Errors vs Messages in samples. 

I would request if you could incorporate these changes within the [MessageDialog.kt](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/blob/main/samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/MessageDialog.kt) file, which by default is used by all compose samples. Granted there may be some refactoring involved.

PR #443, opened 2025-11-19
Minor code & doc cleanup

shubham7109:
## Description

PR to update `v.next` with minor doc & code cleanup for v300 beta.


XuemingRocks (approved):
Changes look good to me. But there is failed formatting check in a ReadMe. Not sure if it needs to be addressed.

eri9000 (approved):

shubham7109:
Thanks for the reviews, to clarify the metata-checker is helpful to align sample-level readmes, the checker here has scanned the root-level repo readme which doesn't need to follow sample-level patterns.

PR #474, opened 2026-04-13
Create a new camera controller on each configuration change

gunt0001:
## Description

Fixes a crash caused by `ObjectAlreadyOwnedException` in sample "Show exploratory viewshed from point in scene" by moving the camera controller out of the view model and declaring it in compose code. This means that a new camera controller is created on every configuraiton change, which means that the "object already owned" issues can be avoided.

https://devtopia.esri.com/runtime/kotlin/issues/7583

## Links and Data

  - [ ] link: https://runtime-kotlin.esri.com/view/Release/job/300.0.0/job/vtest/job/sampleviewer/3/


shubham7109 (approved):
Looks good, thanks @gunt0001!

PR #475, opened 2026-04-13
Alan/applymapalgebra tweaks

alan-edi:
## Description

A few little tweaks to the `apply-map-algebra` sample:
- In the samples-lib `strings.xml`, change the `requires_provisioning` string to match the wording on the DOWNLOAD button. This change will affect all samples that use the `DownloaderActivity`.
- Fix a typo in `README.md` (which was inherited from the sample design README).
- In `ApplyMapAlgebraViewModel`, change a couple of variable names to match the sample design, and modify a few comments.

### DIFF
--- a/samples-lib/src/main/res/values/strings.xml
+++ b/samples-lib/src/main/res/values/strings.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="requires_provisioning">This sample requires local data to be provisioned to the device.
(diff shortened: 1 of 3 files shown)


rom14514 (approved):
@alan-edi Tanks for looking into it! Looks good 👍

alan-edi:
@shubham7109 Please give this its second review. In particular the change to the samples-lib `strings.xml`, since that will affect all samples that use the `DownloaderActivity`.

shubham7109 (approved):

PR #476, opened 2026-04-21
Add support for adaptive samples

shubham7109:
## Description

PR to add support for building adaptive samples.  Introduces a reusable adaptive three-pane scaffold in `samples-lib` and migrates the **Show Exploratory Viewshed From Point In Scene** sample to use it.

## Links and Data
Sample issue: `kotlin/issues/7625`
  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/sampleviewer/258/
  

- Introduces `AdaptiveThreePane` as a reusable adaptive scaffold in `samples-lib` for a consistent three-layer pattern: main content, supporting pane, and optional floating pane.

- Migrates the viewshed sample UI to `AdaptiveThreePane` so layout behavior is responsive across phone, foldable, tablet, and larger widths without per-sample layout duplication.

- Adds pane state behavior in `AdaptiveThreePane` (open/close/toggle + restore anchors) to make supporting pane interactions predictable and easier to reuse in other samples.
 
- Adds floating widget support in `AdaptiveThreePane` with drag + dismiss behavior, so samples can expose quick actions without taking over the full supporting pane.
 
- Updates the sample `MainScreen` to separate state wiring from content rendering, making preview rendering possible without requiring a live `ViewModel` path.

- Refactors sample screen composition to pass callbacks/values into scaffold content instead of tightly coupling UI to `SceneViewModel`, improving testability and previewability.

- Updates viewshed controls to be ViewModel-driven and added reset behavior in `SceneViewModel` to restore viewshed options and animate back to the configured initial camera/viewpoint.

- Adds preview-safe behavior by avoiding ArcGIS runtime object creation in preview composition paths, preventing renderer issues in Compose preview.

- Adds reusable multi-device light/dark preview annotation (`@SampleDeviceLightDarkPreview`) so sample screens can opt into a standard adaptive previews with minimal boilerplate.

### Docs and resources:

- https://developer.android.com/develop/ui/compose/layouts/adaptive/canonical-layouts
- https://developer.android.com/develop/ui/compose/layouts/adaptive/build-a-supporting-pane-layout
- https://m3.material.io/foundations/layout/applying-layout/pane-layouts


samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/adaptive/AdaptiveThreePane.kt
```diff
@@ -0,0 +1,898 @@
...
+                hasUserDraggedSinceLastSettle = false
+                closeSupportingPane()
+                return@LaunchedEffect
+            }
+
+            // Give light haptics when the drag handle settles to a new anchor.
+            if (hasUserDraggedSinceLastSettle && previousSettledAnchorIndex != settledAnchorIndex) {
+                hapticFeedback.performHapticFeedback(HapticFeedbackType.TextHandleMove)
+            }
+            previousSettledAnchorIndex = settledAnchorIndex
```
alan-edi, line 285: AS warns this assigned value never used

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/adaptive/AdaptiveThreePane.kt
```diff
@@ -0,0 +1,898 @@
...
+            lastRestorableAnchorIndex = settledAnchorIndex
+            hasUserDraggedSinceLastSettle = false
+        }
+
+        SupportingPaneScaffold(
```
alan-edi, line 290: AS warns "Calling a androidx.compose.ui.UiComposable composable function where a UI Composable composable was expected" here.

Also several more of this warning below, inside `mainPane` and `supportingPane`, but I'll mention only this one.

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/AdaptiveThreePane.kt
```diff
@@ -0,0 +1,898 @@
...
+
+// Auto-close threshold when supporting pane shrinks to a thin strip.
+private const val WIDE_CLOSE_PROPORTION_THRESHOLD = 0.80f
+
+/**
+ * User-facing knobs for the adaptive three-pane.
+ *
+ * @property supportingPaneInitiallyOpen Whether the supporting pane starts open on first composition.
+ * @property floatingPane Configuration for initial position and drag behavior of the floating pane.
+ * @property compactSupportingPaneHeightRatio Height ratio used for supporting pane when compact-stacked.
```
alan-edi: Doc for `floatingPaneInitiallyVisible` property missing

samples/show-exploratory-viewshed-from-point-in-scene/src/main/java/com/esri/arcgismaps/sample/showexploratoryviewshedfrompointinscene/screens/ViewshedOptionsScreen.kt
```diff
@@ -16,153 +16,167 @@
...
 import androidx.compose.ui.unit.dp
-import com.esri.arcgismaps.sample.sampleslib.theme.SampleAppTheme
+import com.esri.arcgismaps.sample.showexploratoryviewshedfrompointinscene.components.ViewshedUiState
 
-/**
- * Viewshed options screen for sliders and checkboxes
- */
 @Composable
-fun ViewshedOptionsScreen(
+fun ViewshedSlidersContent(
```
alan-edi: Does it still make sense for this file to be called `ViewshedOptionsScreen.kt` now this function has been renamed?

alan-edi (commented):
@shubham7109 Looks very good, and your use of it for the **Show Exploratory Viewshed From Point In Scene** sample is excellent! A few minor comments below. Also, I think:
- In the sample's `README.md`, **How to use the sample** should be updated to mention the Scene Options.
- The screenshot should come from a tablet in landscape orientation, so we can see the sliders alongside the Scene, something like the following (probably needs just a little trimmed off the height to give it a 700x400 aspect ratio):

Image

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/adaptive/AdaptiveThreePane.kt
shubham7109: Would recommend to upgrade to use the latest version of Android Studio, I've seen these before but not anymore on my end and I'm guessing some upgrade may have fixed it. They have been previously mentioned in our PRs, and is currently considered as an Android Studio lint check issue. 

You could verify as the value is being used to updated the anchor index. 

image

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/adaptive/AdaptiveThreePane.kt
shubham7109: Same as above, there shouldn't be warnings on recent AS versions or verify them if they are in-fact true. 

image

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/AdaptiveThreePane.kt
shubham7109: Added, good catch 👍

samples/show-exploratory-viewshed-from-point-in-scene/src/main/java/com/esri/arcgismaps/sample/showexploratoryviewshedfrompointinscene/screens/ViewshedOptionsScreen.kt
shubham7109: Renamed the file and function to opt to use `ViewshedSupportingContent`, and similarly `ViewshedSupportingContent`, to apply general refactoring to place composables based on their content Kotlin files.

I'm taking this as consideration feedback on how to come up with a naming pattern as we would ideally like to use unique file names for new samples. I'm planning on using either pattern:

- SampleNameScreen
- SampleNameSupportingContent 
- SampleNameFloatingContent


OR

- SampleNameScreen
- SampleNameLayoutPanes

We could use these considerations to come up with guidelines as part of the the next steps to update the new module template, but your thoughts would be helpful to have.

shubham7109:
@alan-edi Thanks for your review, I've added the feedback changes. Feel free to give it another look!

samples/show-exploratory-viewshed-from-point-in-scene/src/main/java/com/esri/arcgismaps/sample/showexploratoryviewshedfrompointinscene/screens/MainScreen.kt
```diff
@@ -16,91 +16,130 @@
...
+                maxDistance = 1500f,
+                isFrustumVisible = true,
+                isAnalysisVisible = true,
+            ),
+            mainPaneContent = {},
+        )
+    }
+}
+
+
```
alan-edi: Delete redundant blank line

samples/show-exploratory-viewshed-from-point-in-scene/src/main/java/com/esri/arcgismaps/sample/showexploratoryviewshedfrompointinscene/screens/ViewshedFloatingContent.kt
```diff
@@ -0,0 +1,106 @@
...
+                Text("Align camera with viewshed")
+            }
+            OutlinedButton(onClick = onResetViewshedOptions) {
+                Text("Reset viewshed options")
+            }
+        }
+    }
+}
+
+
```
alan-edi: No need for repeated blank lines

samples/show-exploratory-viewshed-from-point-in-scene/src/main/java/com/esri/arcgismaps/sample/showexploratoryviewshedfrompointinscene/screens/ViewshedSupportingContent.kt
```diff
@@ -0,0 +1,154 @@
+/* Copyright 2023 Esri
```
alan-edi: 2026

samples/show-exploratory-viewshed-from-point-in-scene/src/main/java/com/esri/arcgismaps/sample/showexploratoryviewshedfrompointinscene/screens/ViewshedOptionsScreen.kt
alan-edi: I like the first option, i.e.:
- SampleNameScreen
- SampleNameSupportingContent
- SampleNameFloatingContent

alan-edi (approved):
@shubham7109 Looks great! I've just mentioned a couple of nitpicks.

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/AdaptiveThreePane.kt
```diff
@@ -0,0 +1,905 @@
...
+                )
+            }
+        }
+    }
+}
+
+/**
+ * Simple heading row for the supporting pane.
+ *
+ * Keeps title and close affordance consistent across samples.
```
sorenoid: just curious, what does this mean?

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/AdaptiveThreePane.kt
```diff
@@ -0,0 +1,905 @@
...
+ * Positioning and interaction options for the floating pane.
+ *
+ * @property initialXFraction Initial x offset as a fraction of the host width.
+ * @property initialYFraction Initial y offset as a fraction of the host height.
+ * @property draggable Whether dragging is enabled for the floating pane.
+ */
+data class FloatingPaneConfig(
+    val initialXFraction: Float = 0.05f,
+    val initialYFraction: Float = 0.05f,
+    val draggable: Boolean = true,
```
sorenoid: It doesn't seem like it would ever be desirable for draggable to be false.

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/adaptive/AdaptiveThreePane.kt
sorenoid, line 1: nitpick: this file is a bit too large according to [kotlin coding conventions](https://kotlinlang.org/docs/coding-conventions.html#source-file-organization), maybe the three panes could be broken into three files?

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/adaptive/AdaptiveThreePane.kt
```diff
@@ -0,0 +1,905 @@
...
+}
+
+// Default split anchor used when the user has no prior pane size preference.
+private const val DEFAULT_EXPANSION_ANCHOR_INDEX = 2
+
+// Right-most anchor that acts as a close gesture target.
+private const val AUTO_CLOSE_EXPANSION_ANCHOR_INDEX = 3
+
+// Auto-close threshold when supporting pane shrinks to a thin strip.
+private const val WIDE_CLOSE_PROPORTION_THRESHOLD = 0.80f
```
sorenoid, line 397: imo this is too wide. I would prefer to get very close to the edge before autoclose kicks in, at least in landscape on my Pixel 6 Pro this seemsed like about 15% of the width, where I expect for autoclose to happen at less than 5%.

sorenoid (approved):
This all looks great and works well. I just have some nitpicks. Happy to approve and let you consider my nitpicks...

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/adaptive/AdaptiveThreePane.kt
shubham7109: @sorenoid The constraint to auto-close when then pane is shrunk down to 20% width was somewhat intentional here. If we observe, the supporting pane will expand and shrink based on the width of the pane's layout changes. 

If we were to change to shrink until 5% to auto close, it would mean designing the pane layout to support a Column of that width to render the layout. See the demo below using this setting, to note the layout to shrink until the width is met for auto-close. This may result in un-desired layout wrapping/clipping behavior, which was originally why the close action triggers on 20% pane width. 

Here's a demo using 5% auto close:

https://github.com/user-attachments/assets/42df22c7-1a48-4f8f-88d1-828c9e4574c6

Leaving this currently as is, to use default open anchor for supporting pane, currently is 33% of device width, and auto-close on 20% shrink.

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/adaptive/AdaptiveThreePane.kt
shubham7109: Split this into three files in the `adaptive` package of the samples-lib.

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/AdaptiveThreePane.kt
shubham7109: Removed `draggable` config  👍

samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/AdaptiveThreePane.kt
shubham7109: Removed line 👍

shubham7109:
Thanks for the helpful guidance reviews, merging this in!

PR #481, opened 2026-05-18
Patch metadata tools script

shubham7109:
## Description

PR to patch up, update and remove legacy docs of the metadata script from the runtime SDK era.


shubham7109:
Note: Git checker is failing, as the updated readme is an unexpected non-sample readme.

PR #489, opened 2026-06-23
showlineofsightanalysisinmap: update bullets in README.md

alan-edi:
## Description

Just updates a bulleted list in `show-line-of-sight-analysis-in-map/README.md`. (When verifying this new sample I noticed the list didn't render correctly when viewed in the Sample Viewer, even though it rendered fine in Github.)

I copied the technique from `add-feature-layers/README.md`.

## What To Review

-  Just check that list in the README file.

### DIFF
--- a/samples/show-line-of-sight-analysis-in-map/README.md
+++ b/samples/show-line-of-sight-analysis-in-map/README.md
@@ -18,9 +18,9 @@ The sample loads with a map centered on the Isle of Arran, Scotland, and runs a 
 
 1. Create an `ArcGISMap` and set it on a `MapView`.
 2. Create three `GraphicsOverlay`s as follows:
-    - One overlay to hold the target's `Graphic`.
-    - One overlay to display a `Graphic` for each observer. Observers are in a separate `GraphicsOverlay` to allow us to detect when an observer graphic is tapped.
-    - One overlay to display the line of sight result graphics.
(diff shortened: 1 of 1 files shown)


alan-edi:
@hud10837 @sorenoid Could just one of you review this tiny change today please, so I can merge today and verify/close this long-standing issue tomorrow?

hud10837 (approved):
Looks good to me. Just for reference, would `-` also work?

alan-edi:
@hud10837 I've tried `-` and that failed, see below for details. But I also found `*` works if it's preceded by 3 spaces so I've changed to doing that.

Full explanation:
1. The original problem was that, only when viewed in the Sample Viewer, numbering of the list reset to 1 for the first item after the sublist.
2. The technique I copied from `add-feature-layers/README.md` solved this, but seemed weird because it only worked when the lines immediately preceding the sublist items had 2 trailing spaces!
3. Your question re `-` prompted me to investigate further. I found that `-` and `*` worked if they were preceded by 3 spaces, whereas I was initially using `*` preceded by just 2 spaces.
4. But `-` caused the automatic README format checker to fail with `MD004 Unordered list style`. I don't know why, but guess it may be caused by `*` being used elsewhere in the document.
5. This has led me to favor `*` preceded by 3 spaces.

hud10837 (approved):
Strange behavior, but if it works I think the `*` renders better

PR #494, opened 2026-06-30
NewModuleScript: Adaptive template support

shubham7109:
## Description

PR to enhance the new module script to provide a new template option to use the [AdaptiveThreePane](https://github.com/Esri/arcgis-maps-sdk-kotlin-samples/blob/v.next/samples-lib/src/main/java/com/esri/arcgismaps/sample/sampleslib/components/adaptive/AdaptiveThreePane.kt) to generate a starting sample template. 

```console
% java -jar tools/NewModuleScript.jar 
Enter Name of the sample with spaces (Eg. "Display new map"): 
Display new adaptive map
Choose the sample category: 
1:  Analysis 
2:  Augmented Reality 
3:  Cloud and Portal 
4:  Layers 
5:  Edit and Manage Data 
6:  Maps 
7:  Scenes 
8:  Routing and Logistics 
9:  Utility Networks 
10: Search and Query 
11: Visualization 
12: Accessibility
Enter a number (1-12) to sample category: 12
Choose the sample template type: 
1: Basic 
2: FAB with DialogOptions 
3: FAB with BottomSheetOptions 
4: Adaptive panes
Enter a number (1-4) to sample template type: 4
Using repository... /arcgis-maps-sdk-kotlin-samples/
Sample Successfully Created!

```

## Links and Data

Sample issue: `runtime/kotlin/issues/7637`
  - [x] link: N/A

## How to test

- Run the script `java -jar tools/NewModuleScript.jar `
- Use any sample name: `Display new adaptive map`
- Choose `4` Adaptive panes template type
- Run sample and review generated template code

## What to review

- The intention is to isolate ArcGIS states vs UI state hoisting through explicit patterns. Any feedback here is appreciated.
- Does the adaptive template provide guidance to best meet Compose & ViewModel patterns?
- Any additional guidance the template should offer sample authors?


kaushikrw (approved):
It works as expected 👍

tools/NewModuleScript/templates/adaptive/MapViewModelTemplate.kt
```diff
@@ -0,0 +1,129 @@
...
+
+class MapViewModel(app: Application) : AndroidViewModel(app) {
+
+    // TODO - Use mutable state if the map instance changes to allow screen to observe updates.
+    val arcGISMap = ArcGISMap(spatialReference = SpatialReference.webMercator()).apply {
+        initialViewpoint = Viewpoint(34.05, -117.19, 2e6)
+    }
+
+    // Create a state flow to hold the UI state for the supporting pane controls
+    private val _adaptiveUiState = MutableStateFlow(AdaptiveUiState.defaultState)
```
hud10837, line 41: In the other template, this is just a `MutableState`. Why do you use a flow here?

tools/NewModuleScript/templates/adaptive/MapViewModelTemplate.kt
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * TODO: Use Kotlin extensions for types to simplify sample workflow.
+ */
+fun BasemapOptions.getBasemapStyle(): BasemapStyle {
+    return when (this) {
+        BasemapOptions.Light -> BasemapStyle.ArcGISLightGray
+        BasemapOptions.Dark -> BasemapStyle.ArcGISDarkGray
+    }
+}
```
hud10837, line 129: This seems to me like it's overcomplicating a very simple sample. `BasemapStyle.ArcGISLightGray` is pretty self-explanatory, I think, and I don't feel like `BasemapOptions.Light` adds much value.

tools/NewModuleScript/templates/adaptive/MapViewModelTemplate.kt
```diff
@@ -0,0 +1,129 @@
...
+    val isLayersEnabled: Boolean
+) {
+    companion object {
+        val defaultState = AdaptiveUiState(
+            basemapOptions = BasemapOptions.Light,
+            isLayersEnabled = true
+        )
+
+    }
+}
```
hud10837, line 112: What is the benefit of bundling this into a class? I can see a benefit in a 1-1 correspondence between a static state object and a resulting UI. However, this adds a lot of complexity to the template which is otherwise quite simple. My understanding was that the point of a state object was to simplify scenarios where you need to read/write interconnected data. For example, a `TextFieldState` lets you easily manage setting text and setting the highlighted area and the cursor position [all at once](https://developer.android.com/reference/kotlin/androidx/compose/foundation/text/input/TextFieldState#(androidx.compose.foundation.text.input.TextFieldState).setTextAndPlaceCursorAtEnd(kotlin.String)). Since these things happen simultaneously, there is room for errors to occur when things are not correctly updated.

However, `TextFieldState` is an inherently mutable type. `AdaptiveUiState` here is immutable and used immutably. If we bundle all of the immutable data together, then wouldn't the composables that call this get recomposed when the state object changes rather than when the individual components change, thus causing more recompositions than necessary?

tools/NewModuleScript/templates/adaptive/SupportingPaneTemplate.kt
```diff
@@ -0,0 +1,163 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import com.esri.arcgismaps.sample.displaycomposablemapview.components.AdaptiveUiState
+import com.esri.arcgismaps.sample.displaycomposablemapview.components.BasemapOptions
+
+/**
+ * Supporting pane content for the sample.
+ */
+@Composable
+internal fun DisplayComposableMapViewSupportingPane(
```
hud10837, line 43: Should this not be a noun?
```suggestion
internal fun ComposableMapViewSupportingPane(
```

hud10837 (changes requested):
The new script works for me.

> The intention is to isolate ArcGIS states vs UI state hoisting through explicit patterns. Any feedback here is appreciated.

I feel that in the template sample, the patterns you introduced complicate the logic of the viewmodel, and the separation between "ArcGIS" states and "UI" state is not so clear. Actually, I usually would think of the basemap and layers being enabled/disabled as a core part of the app UI rather than business logic related to app functionality that we expose as part of the SDK. What do you think?

tools/NewModuleScript/templates/adaptive/MapViewModelTemplate.kt
shubham7109: The flow is mainly for providing ease of hoisting UI to collect state with lifecycle of the entry composable. This also allows for easy preview functionality, I add more details why in the comment below.


Using a `MutableState` as well would need some support to drive the _initial_ state of the sample and its controls, so either would work fine. This template chooses a `MutableStateFlow` like many samples have for exposing UI controls which currently the templates do not.

tools/NewModuleScript/templates/adaptive/MapViewModelTemplate.kt
shubham7109: > Use custom enums / data class to provide ease of UI options.

This choice of design is primarily intended to enable using [Interactive mode](https://developer.android.com/develop/ui/compose/tooling/previews#preview-interactive) with Compose previews, to encourage using Kotlin types that are preview friendly enabling live preview of sample UI. This is possible by explicitly isolating the ArcGIS objects to live within the ViewModel and use Kotlin objects to live within the composable screen. 

This allows for clear separation of logic for UI vs the SDK, as the screen may also need a displayed name/state associated with the view model and not all SDK objects need to be exposed for UI.  Ideally yes, samples should prioritize SDK clarity and avoid unnecessary architecture/UI complexity, see how this template showcases using an enum to iterate and build the composable UI, `entries.forEach { option ->`, then to display the `option.name` or highlight the selected option using simple Kotlin assertions. Keeping all ArcGIS Maps SDK usage within the view model file.

tools/NewModuleScript/templates/adaptive/MapViewModelTemplate.kt
shubham7109: I see the concern here, `TextFieldState` may be a different case because it models tightly coupled mutable UI element state, while `AdaptiveUiState` for the template is meant as a immutable screen state.

The reason I grouped UI states is to follow Compose state-hoisting [guidance](https://developer.android.com/develop/ui/compose/state) for screen UI state: 
>When hoisting state, there are three rules to help you figure out where state should go:
>1. State should be hoisted to at least the lowest common parent of all composables that use the state (read).
>2. State should be hoisted to at least the highest level it may be changed (write).
>3. If two states change in response to the same events they should be hoisted together.

The template provided showcases the ViewModel object which owns the UI state, the top-level screen collects it, allowing the scaffold stays stateless and previewable. That gives unidirectional data flow for compose.

On recomposition, using a data class is not should not inherently be an issue. Compose recomposes based on state reads and can skip unaffected parts of the tree. So object grouping alone is not a risk, but streamlines the extrapolation for the screen. That said, if we want the template to stay as minimal as possible, flattening the state surface would also be a reasonable direction, to note the current (3) templates provides a direct state managed updates from the view model.

> there is room for errors to occur when things are not correctly updated.

I do agree there is a tradeoff. If these values are not actually coupled, a bundled UiState can feel like extra structure for a template. So, I would advice this pattern to be leveraged mostly for complex or UI heavy samples, as this pattern is useful for consistency, previewability, and scaling for sample controls

Let me know if guidance is to change the pattern, for reference Compose provides docs for [UI state and business logic](https://developer.android.com/develop/ui/compose/state-hoisting#types-of-state-and-logic)

tools/NewModuleScript/templates/adaptive/SupportingPaneTemplate.kt
shubham7109: This is a placeholder function name for the script, which will be renamed to match the generated `.kt` file, using the sample name, e.g. `DisplayAdaptiveSampleSupportingPane(...)`. 

This pattern is similarly used in the other (3) templates while generating a sample, and also would encourage authors choose an appropriate name as needed. Here the motive is mainly to use unique fun names, to aim to simplify debugging through stack-traces and avoiding conflicting with other sample function names.

hud10837 (approved):
Thank you for the thorough explanations

PR #522, opened 2026-09-02
Dynamic base map gallery

betterForFun:
## Description
PR to add a new Kotlin sample _"SAMPLE_NAME"_ in `SAMPLE_CATEGORY` category.

## Links and Data
Sample Epic: `runtime/kotlin/issues/4572`
  - [x] link: https://runtime-kotlin.esri.com/view/all/job/vtest/job/sampleviewer/312/


colinanderson:
@betterForFun you need to update the screen shot to something that shows the app.

The design looks like it is saying to not use the toolkit basemap gallery 
>This sample demonstrates how to implement a basemap gallery using the Maps SDK. The styles and associated customization options used for the gallery are retrieved from the [basemap styles service](https://developers.arcgis.com/rest/basemap-styles/). A ready-made basemap gallery component is also available in the toolkit's provided with each SDK. To see how the ready-made basemap gallery toolkit component can be integrated into a Maps SDK application refer to the Set Basemap sample.

So I assume you are using the toolkit here because the Kotlin basemap gallery is not coupled to a geomodel so is flexible enough to provide the custom gallery that this sample needs?

colinanderson:
Some questions/observations
- I think the currently visible basemap should be selected in the gallery
- The drop down menus don't work when no basemap is selected - maybe disable them in this case. Auto selecting the current basemap probably fixes this
- Once I have selected a basemap I can't hit done without changing at least one dropdown - so "Default" for both is invalid
- Looks like I can't proceed without choosing a language - should world view be disabled until I choose a language?

samples/dynamic-basemap-gallery/src/main/java/com/esri/arcgismaps/sample/dynamicbasemapgallery/MainActivity.kt
```diff
@@ -0,0 +1,43 @@
...
+
+package com.esri.arcgismaps.sample.dynamicbasemapgallery
+
+import android.os.Bundle
+import androidx.activity.ComponentActivity
+import androidx.activity.compose.setContent
+import androidx.activity.enableEdgeToEdge
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Surface
+import androidx.compose.runtime.Composable
```
colinanderson: This import is unused

samples/dynamic-basemap-gallery/src/main/java/com/esri/arcgismaps/sample/dynamicbasemapgallery/MainActivity.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.esri.arcgismaps.sample.sampleslib.theme.SampleAppTheme
+import com.esri.arcgismaps.sample.dynamicbasemapgallery.screens.DynamicBasemapGalleryScreen
+import com.esri.arcgismaps.sample.sampleslib.theme.SampleAppTheme
```
colinanderson: This import is unused

samples/dynamic-basemap-gallery/src/main/java/com/esri/arcgismaps/sample/dynamicbasemapgallery/screens/DynamicBasemapGalleryScreen.kt
```diff
@@ -0,0 +1,209 @@
...
+import com.esri.arcgismaps.sample.sampleslib.components.SampleTopAppBar
+
+/**
+ * Main screen layout for the sample app
+ */
+@Composable
+fun DynamicBasemapGalleryScreen(
+    mapViewModel: DynamicBasemapGalleryViewModel = viewModel()
+) {
+    val mapViewModel: DynamicBasemapGalleryViewModel = viewModel()
```
colinanderson: I think you can delete this line because the default `mapViewModel` argument is the same value.

betterForFun:
@colinanderson thank you Colin, both comments are resolved

colinanderson:
>So I assume you are using the toolkit here because the Kotlin basemap gallery is not coupled to a geomodel so is flexible enough to provide the custom gallery that this sample needs?

This wasn't a suggestion to remove using the basemap gallery just checking that no one had specifically said not to use it. I think it is fine to use because it should be flexible enough to work in this sample.

Also, with your implementation the selected basemap can be hard to see especially if it is a dark map. At first I didn't even realise there was a sort of highlight and assumed selection was only shown by the colour of the font.

Basemap Gallery
image

Sample
image

samples/dynamic-basemap-gallery/src/main/java/com/esri/arcgismaps/sample/dynamicbasemapgallery/MainActivity.kt
colinanderson:
```
import androidx.compose.runtime.Composable
import com.esri.arcgismaps.sample.sampleslib.theme.SampleAppTheme
```
are showing as unused for me

samples/dynamic-basemap-gallery/src/main/java/com/esri/arcgismaps/sample/dynamicbasemapgallery/screens/DynamicBasemapGalleryScreen.kt
```diff
@@ -0,0 +1,290 @@
...
+import com.esri.arcgismaps.sample.sampleslib.components.DropDownMenuBox
+import com.esri.arcgismaps.sample.sampleslib.components.MessageDialog
+import com.esri.arcgismaps.sample.sampleslib.components.SampleTopAppBar
+
+/**
+ * Main screen layout for the sample app
+ */
+@Composable
+fun DynamicBasemapGalleryScreen(
+    mapViewModel: DynamicBasemapGalleryViewModel = viewModel()
```
colinanderson: I think you should remove the view model parameter. As far as I remember samples usually just create the viewmodel as you do on line 79 below.

Also, having the viewmodel as a param is flagged as unstable.

samples/dynamic-basemap-gallery/src/main/java/com/esri/arcgismaps/sample/dynamicbasemapgallery/screens/DynamicBasemapGalleryScreen.kt
```diff
@@ -0,0 +1,290 @@
...
+                )
+            }
+        }
+
+        // Display menu
+        if (showBasemapGallery) {
+            // Default the pending selection to the actual gallery item for the basemap style
+            // currently applied to the map, so it can be highlighted when the gallery opens
+            var pendingItem by remember {
+                mutableStateOf(
```
colinanderson, line 117: Do the pending values actually need to be remembered? Shouldn't they be reset from the view model each time the dialog shows since the view model has the true value?

As far as I can see the current code works but we have two values in play - pending and view model - so there could be a way that they get out of sync.

I noticed if I open the dialog before the basemap has appeared that the dialog then doesn't reflect the displayed basemap so it is not selected - I wonder if that is caused by remembering `null` before the model has loaded its basemap.

https://github.com/user-attachments/assets/77d51540-0785-4916-bcf3-72708d35414a

samples/dynamic-basemap-gallery/README.md
```diff
@@ -0,0 +1,33 @@
+# Dynamic base map gallery
+
+Dynamic base map gallery
+
+![Image of cut geometry](cut-geometry.png)
```
colinanderson: This looks wrong. Shouldn't it be
```
![Image of dynamic basemap gallery](dynamic-basemap-gallery.png)
```

samples/dynamic-basemap-gallery/README.metadata.json
```diff
@@ -0,0 +1,24 @@
...
+        "dynamic-basemap-gallery.png"
+    ],
+    "keywords": [
+        "basemapgallery map",
+        "basemapgallery"
+    ],
+    "language": "kotlin",
+    "redirect_from": "TODO",
+    "relevant_apis": [
+        "basemapgallery"
```
colinanderson, line 18: I think this might be wrong if you are not using the toolkit basemap gallery

colinanderson (commented):
@betterForFun looks good. A few comments to look over.

samples/dynamic-basemap-gallery/src/main/java/com/esri/arcgismaps/sample/dynamicbasemapgallery/screens/DynamicBasemapGalleryScreen.kt
betterForFun: Hi Colin, I think the pending items is just a place holder for the dialog display, it does not sync with viewmodel unless the actual style is changed, android requires the mutable var to be included inside "remember"

samples/dynamic-basemap-gallery/README.metadata.json
betterForFun: This is auto-generated. It fails the meta check if I remove it

samples/dynamic-basemap-gallery/src/main/java/com/esri/arcgismaps/sample/dynamicbasemapgallery/screens/DynamicBasemapGalleryScreen.kt
colinanderson: Yes, but the pending value is remembering sate from the view model so in theory the UI can become out of sync with the view model.

Would it be better to put the pending values into the view model with a function to apply them called when the dialog is applied? They could be mutable state in the view model.

Either way you need to address the bug shown above where the dialog is populated before the map view appears and therefore doesn't reflect the map view state.

colinanderson: Ah, ok looks like maybe you fixed the bug with the key. For the future if you've addressed a review comment like that add a comment saying that you did. Just makes life easier for the reviewer.

colinanderson: I still think it might be better to put the pending values into the view model though.

samples/dynamic-basemap-gallery/src/main/java/com/esri/arcgismaps/sample/dynamicbasemapgallery/screens/DynamicBasemapGalleryScreen.kt
```diff
@@ -0,0 +1,284 @@
...
+            // Keyed on selectedBasemapStyleInfo so it re-syncs once the basemap style service
+            // finishes loading, in case the gallery was opened before that completed.
+            var pendingItem by remember(mapViewModel.selectedBasemapStyleInfo) {
+                mutableStateOf(
+                    mapViewModel.basemapGalleryItems.firstOrNull { item ->
+                        (item.tag as? BasemapStyleInfo)?.style == mapViewModel.selectedBasemapStyleInfo?.style
+                    }
+                )
+            }
+            var pendingLanguage by remember { mutableStateOf(mapViewModel.selectedLanguage) }
```
colinanderson: Shouldn't these two be keyed as well?

samples/dynamic-basemap-gallery/src/main/java/com/esri/arcgismaps/sample/dynamicbasemapgallery/screens/DynamicBasemapGalleryScreen.kt
betterForFun: > Ah, ok looks like maybe you fixed the bug with the key. For the future if you've addressed a review comment like that add a comment saying that you did. Just makes life easier for the reviewer.

Got it! Sorry for the confusion, I'll make sure to note that in the future

colinanderson (approved):
@betterForFun looks like there's just keying or not of the other pending values to be decided so I'll approve to avoid further delays.

betterForFun:
> @betterForFun looks like there's just keying or not of the other pending values to be decided so I'll approve to avoid further delays.

Thank you Colin, I'll fix the keys and do some research on state hoisting

sorenoid (approved):

